PR #1, opened 2025-11-04
Unplaced Reconfigurable GEMM, Accuracy Flag Refactor, Add Throughput Metric

kurtis-b-1:
The current GEMM is placed, while the rest of the examples are unplaced. Moreover, it's hard to determine how the `prio-accuracy` feature affects the execution of GEMM. Lastly, no metric for throughput is saved for GEMM. This PR tries to address the first and third points and improve the second point. 

Let me know if I should split this up into multiple PRs to make it easier to review. 

## Added
-Flag in `CMakeLists.txt` to enable/disable `prio-accuracy` in GEMM, with the appropriate kernel used based on whether this flag is enabled or disabled
-Throughput calculation and metric added to CI in GFLOP/s

## Changed
-Placed reconfigurable GEMM design is now unplaced reconfigurable GEMM
-Clearer blocking regarding how the `prio-accuracy` flag changes the design 

## Removed
-Placed reconfigurable GEMM design

## PR Merge Checklist

### DIFF
--- a/applications/llama_3.2_1b/src/operator/aie_gemm.py
+++ b/applications/llama_3.2_1b/src/operator/aie_gemm.py
@@ -121,7 +121,7 @@ class AIEGEMM(AIEOperatorBase):
                 "trace_size": 0,
                 "generate_taps": False,
             },
-            requires_context=True,
+            requires_context=False,
         )
 
         # FIXME: We should be able to reuse the same xclbin for same tile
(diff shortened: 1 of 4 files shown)


andrej (approved):
Looks good to me.

To fix the llama error, you might have to change this to False:
https://github.com/amd/IRON/blob/aec88fd7782c9f5aa88f82cd588ad19b89335435/applications/llama_3.2_1b/src/operator/aie_gemm.py#L118

Not directly related to this PR, but we should probably also start testing other types again. In IRON(clad) it's been only `bf16`, and I'd assume since we made some changes to this and never tested other types, we probably already broke support for other types. If the goal is for this to become the main matmul implementation and removing all but the getting started example from MLIR-AIE, that support will need to be added back. cc @jgmelber @pvasireddy-amd

kurtis-b-1:
> Looks good to me.
> 
> To fix the llama error, you might have to change this to False:
> 
> https://github.com/amd/IRON/blob/aec88fd7782c9f5aa88f82cd588ad19b89335435/applications/llama_3.2_1b/src/operator/aie_gemm.py#L118

Yup to run inference with these changes, I had to set that flag to False

PR #8, opened 2025-11-13
Add Example Application: Llama 3.2 (1 billion parameters)

andrej:
## Added
- NPU-offloaded implementation of Llama 3.2 (1 billion parameters)
- New CI test for llama

## Changed
- Moved `example` -> `operators`
- Created `examples` directory to hold llama
- Removed reliance on environment variables for MLIR AIE and Peano directories, since we no longer use an env setup script

## PR Merge Checklist

1. [x] The PR is rebased on the latest `devel` commit and pointing to `devel`.
2. [x] Your PR reviewed and approved.
3. [x] All checks are passing.

PR #10, opened 2025-11-13
Llama: Keep weights stationary in dedicated buffers

andrej:
Builds on #8.

## Changed
- Weights get dedicated buffer allocations for each operator, such that weights do not need to be copied on each operator invocation.

## PR Merge Checklist

### DIFF
--- a/applications/llama_3.2_1b/src/block/feed_forward.py
+++ b/applications/llama_3.2_1b/src/block/feed_forward.py
@@ -63,6 +63,16 @@ class FeedForward(nn.Module):
             self.fc3 = AIEGEMM(
                 M=M_prefill, K=self.hidden_dim, N=self.emb_dim, **aie_config_prefill
             )
+        else:
+            self.fc1 = nn.Linear(
+                cfg["emb_dim"], cfg["hidden_dim"], dtype=cfg["dtype"], bias=False
+            )
+            self.fc2 = nn.Linear(
+                cfg["emb_dim"], cfg["hidden_dim"], dtype=cfg["dtype"], bias=False
+            )
+            self.fc3 = nn.Linear(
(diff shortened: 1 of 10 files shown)


andrej:
Changed base to PR #8 to make the diff smaller and review easier. The idea is to merge #8 first, then merge this into `devel`.

applications/llama_3.2_1b/src/block/feed_forward.py
```diff
@@ -112,15 +112,8 @@ def forward(self, x):
         is_prefill = not is_vector or not self.cfg["use_kv_cache"]
```
kurtis-b-1: I think `is_prefill` isn't used?

applications/llama_3.2_1b/src/block/feed_forward.py
```diff
@@ -112,15 +112,8 @@ def forward(self, x):
...
 
         if is_vector and self.cfg["use_kv_cache"] and self.cfg["use_aie_gemv"]:
```
kurtis-b-1: The first two conditions here are the same as `is_decode_with_kv`. Is it fine to just replace them--I think it would be here and a few lines below too.

kurtis-b-1 (approved):
LGTM! A couple minor comments, but they can be in a different PR dedicated to cleanup.

applications/llama_3.2_1b/src/block/feed_forward.py
andrej: good catch

applications/llama_3.2_1b/src/block/feed_forward.py
andrej: good catch

PR #11, opened 2025-11-13
Llama: Runlist-based SwiGLU

andrej:
Builds on #10.

## Added
- Introduce runlist-based SwiGLU

## Changed
- Use XRT runlist for `run_runlist` in AIE operations
- Small fix to compilation steps

## PR Merge Checklist

---

## To Do

 - [x] Address Comment from Curt on original PR in internal repo:
   > Would it be better to have an abstract get_artifacts() method in the base class that child classes have to override?
 - [x] Double check why output looks slightly different after this -- changes to RNG were the root cause of this
 - [x] Double check all combinations of configuration options still work


andrej:
Changed base to PR #10 to make the diff smaller and review easier. The idea is to merge #10 first, then merge this into `devel`.

applications/llama_3.2_1b/src/block/feed_forward.py
```diff
@@ -6,13 +6,17 @@
 # SPDX-FileCopyrightText: Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import logging
```
kurtis-b-1: Is logging necessary, or was it just a leftover from something?

applications/llama_3.2_1b/src/block/feed_forward.py
```diff
@@ -25,6 +29,16 @@ def __init__(
...
         self.cfg = cfg.copy()
 
+        assert (
+            cfg["use_aie_ffn_swiglu"]
+            and not (
+                cfg["use_aie_ffn_silu"]
+                or cfg["use_aie_ffn_gemm"]
+                or cfg["use_aie_ffn_mul"]
+            )
+            or not cfg["use_aie_ffn_swiglu"]
```
kurtis-b-1, line 38: If you have `cfg["use_aie_ffn_swiglu"] or not cfg["use_aie_ffn_swiglu"]` as part of the conditions, wouldn't the assertion never fail?

applications/llama_3.2_1b/src/operator/aie_elementwise_mul.py
```diff
@@ -15,6 +15,49 @@
...
+        f"{file_name_base}.mlir",
+        import_path=base_dir / "example" / "elementwise_mul" / "eltwise_mul.py",
+        callback_fn="my_eltwise_mul",
+        callback_args=[
+            device_type,
+            size,
+            num_columns,
+            num_channels,
+            tile_size,
+            0,
```
kurtis-b-1: I think this is trace size right? It might be better to add it as another argument with default 0.

applications/llama_3.2_1b/src/operator/aie_swiglu_decode.py
```diff
@@ -0,0 +1,171 @@
...
+    SourceArtifact,
+    PythonGeneratedMLIRArtifact,
+)
+from ..utils import torch_to_numpy, numpy_to_torch
+from .aie_gemv import get_gemv_artifacts
+from .aie_silu import get_silu_artifacts
+from .aie_elementwise_mul import get_elementwise_mul_artifacts
+
+
+class AIESwiGLUDecode(AIEOperatorBase):
```
kurtis-b-1, line 25: Probably for a future PR, but I think it would be better to have a single `AIESwiGLU` class that has prefill and decode modes.

applications/llama_3.2_1b/src/operator/aie_gemm.py
```diff
@@ -18,6 +18,107 @@
 from ..utils import torch_to_numpy, numpy_to_torch
 
 
+# Factoring out the GEMM artifact definition because it is reused as a component of other operators (SwiGLU)
```
kurtis-b-1: I think we discussed this before, but it might be nicer to create a `get_artifacts()` method inside of the base class that child classes have to override. Because if someone wants to create another bigger operator similar to SwiGLU that uses other operators as components, then they would have to define operator-specific `get_[]_artifacts()` functions if they're not present.

kurtis-b-1 (approved):
Had some minor comments and remarks about potential future PRs. But overall looks good to me.

applications/llama_3.2_1b/src/block/feed_forward.py
andrej: Indeed, `A or not A` is a tautology, but `(A and B) or not A` is not, which is what this is. `and` binds stronger than `or`. This is just another way of expressing an implication `A -> B` which I think Python doesn't have.

applications/llama_3.2_1b/src/block/feed_forward.py
andrej: just a  leftover! removed

applications/llama_3.2_1b/src/block/feed_forward.py
kurtis-b-1: Agreed I was thinking too simply. But I think it can be expressed as `A.B+A'`, which can be distributed to be 
`(A'+A).(A'+B)`. This would simplify to `1.(A'+B)=A'+B`. 
Since `B` is really more like `(C+D+E)'`, then you can do `(A.(C+D+E))'`, which is something like
`assert (
            not ((
                cfg["use_aie_ffn_silu"]
                or cfg["use_aie_ffn_gemm"]
                or cfg["use_aie_ffn_mul"]
            )
            and cfg["use_aie_ffn_swiglu"])`

EDIT: Sorry I don't mean to be particular about this. I just got excited about boolean algebra lol.

applications/llama_3.2_1b/src/operator/aie_gemm.py
andrej: Done!

applications/llama_3.2_1b/src/operator/aie_swiglu_decode.py
andrej: Leaving this for a future PR because it would involve having multiple different runlists for one operator. Currently each operator has exactly one runlist. We should discuss this.

applications/llama_3.2_1b/src/operator/aie_elementwise_mul.py
andrej: done

applications/llama_3.2_1b/src/block/feed_forward.py
andrej: Haha don't ever apologize for getting too excited about boolean algebra. Not sure your expression actually is equivalent though. What would happen if I enabled, e.g. "use_aie_ffn_silu", but none of the others. That should be allowed, but your assert would be False, I think.

kurtis-b-1: No, the assert would be True--the first `not` is wrapped around the full conditional statement. So, in an example where only `use_aie_ffn_silu=True` it would evaluate to `((1+0+0).0)'=1`.

andrej: Aaah that makes sense. I didn't see that parentheses. Somehow I find mine easier to understand though :p

kurtis-b-1: To each their own!

PR #14, opened 2025-11-13
TTFT Performance Improvements

kurtis-b-1:
Immediate optimization opportunities were found from profiling. This PR incorporates these optimizations.

Builds upon https://github.com/amd/IRON/pull/8. 

## Added
- Missing microkernel dims and kernel flags for AIE GEMM operator

## Changed
- Pipelined rms norm and eltwise mul for weighted rms norm operation
- AIE operators refactored based on the Element-Wise Addition AIE operator as a template: SiLU, Element-Wise Mul, RMS Norm
- GEMM kernel compilation changes based on whether prio_accuracy flag is True/False

## Removed
-

## PR Merge Checklist


andrej:
To make review easier (smaller diff, exclude changes from adding llama), you can change the base of this PR to PR #8 (branch `bring-in-the-llama`). Took me a sec to find this (you might already know how) but it's if you click "Edit" at the top of the page, then under the title text box select a different base. We'll merge #8 first, and I think then the base of this should automatically change to `devel`.  I've just done the same for #10 and #11.

PR #16, opened 2025-11-14
Fix Task Group Mixing

kurtis-b-1:
Address https://github.com/amd/IRON/issues/15. 

## Added
- Using task groups in the inner-most loop of MHA runtime. Without it, a long list of BD errors shows up. Checking with previous CI results, this change doesn't seem to negatively affect the performance of MHA.

## Changed
- Use same task groups for fill operations as the drain operations in operators such as AXPY, etc. 

## PR Merge Checklist


andrej (approved):
Thanks for fixing this so quickly

PR #19, opened 2025-11-17
Llama Custom Profiler

kurtis-b-1:
Custom profiler for performance measurements of Llama 3.2-1B inference. A logging function is executed at the call and return of each function. Initially tried with `cProfile`, but the measurements of PyXRT execution was missing (possibly `cProfile` excludes I/O like the NPU). This method allows for deterministic profiling of inference, though profiling overhead is higher compared to statistical profiler. But this method is apt for getting a comprehensive picture of the inference run, as one can check the time of each AIE and CPU operator relative to the inference metrics such as TTFT. The overhead of profiling shouldn't skew the relative contribution of each operator to the latency of inference.

## Added
- Script for generating CSV based on log

## Changed
- Attach custom profiler to system during inference if a CLI argument is passed

## Removed
-

## PR Merge Checklist


applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+import argparse
+import re
+from collections import defaultdict
+from pathlib import Path
+import sys
+import csv
+
+
+class FunctionStats:
+    """Store statistics for a function."""
```
andrej: Comment seems redundant, the name of the class tells me that.

applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+    Parse a profile log file and extract function timing information.
+
+    Args:
+        log_file: Path to the profile log file
+
+    Returns:
+        dict: Dictionary mapping function names to FunctionStats objects
+    """
+    stats = defaultdict(lambda: FunctionStats(""))
+    call_stack = {}  # Track ongoing calls by function identifier
```
andrej: I find this variable name misleading. This is not a call stack in the usual sense where you push and pop for each function call/return. It appears you use this variable to keep track of only the _most recent_ starting time of a function call of that name. 

Also, are you actually using the values that you write to that dict anywhere? I think this may be a leftover from an earlier implementation that could be removed.

applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+
+    with open(log_file, "r") as f:
+        for line in f:
+            # Try to match CALL pattern
+            call_match = call_pattern.search(line)
+            if call_match:
+                func_id = call_match.group(1)
+                timestamp = float(call_match.group(2))
+                call_stack[func_id] = timestamp
+                continue
```
andrej, line 75: I think this is unused

applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+            print(f"  ... and {len(stats) - 20} more")
+        return
+
+    print("\n" + "=" * 120)
+    print(f"DETAILED STATISTICS FOR FUNCTIONS CONTAINING: '{function_name}'")
+    print(f"Found {len(matching_funcs)} matching function(s)")
+    print("=" * 120)
+
+    import statistics
+
```
andrej: Move import to the top of the file please

applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+    def median_time(self):
+        """Calculate median time."""
+        if not self.durations:
+            return 0.0
+        sorted_durations = sorted(self.durations)
+        n = len(sorted_durations)
+        if n % 2 == 0:
+            return (sorted_durations[n // 2 - 1] + sorted_durations[n // 2]) / 2
+        else:
+            return sorted_durations[n // 2]
```
andrej: You import statistics in the file later. I suggest using the standard implementation of median since you're already importing it

applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+
+    args = parser.parse_args()
+
+    # Determine log file to analyze
+    if args.log_file:
+        log_file = Path(args.log_file)
+    else:
+        # Find most recent profile log
+        logs_dir = Path("logs")
+        if not logs_dir.exists():
```
andrej: No need to fix this in this script since this isn't critical, but just sharing something you might find interesting so you know: Using a "file exists" check followed by then opening that file is almost always not what you want. The reason is time-of-check-vs.-time-of-use (TOCTOU) issues. The file might be deleted right after you check and before you open it, so the check really doesn't tell you anything useful. The preferred approach is to just try to open the file; if you succeed, you can use it, if you fail, you can catch the exception and let the user know the file doesn't exist.

applications/llama_3.2_1b/inference.py
```diff
@@ -29,6 +29,98 @@
     generate,
 )
 
+# Global dictionary to track function call times
+_function_call_stack = []
```
andrej: That's a list/stack (of dictionaries) :)

applications/llama_3.2_1b/analyze_profile.py
```diff
@@ -0,0 +1,400 @@
...
+            f"{func_stats.total_time:>12.6f} {func_stats.avg_time:>12.6f} "
+            f"{func_stats.min_time:>12.6f} {func_stats.max_time:>12.6f} "
+            f"{func_stats.median_time:>12.6f}"
+        )
+
+    print("-" * 160)
+
+    # Print overall statistics
+    total_calls = sum(s.call_count for s in filtered_stats.values())
+    total_time = sum(s.total_time for s in filtered_stats.values())
```
andrej: Doesn't this double-count durations since the function logs its execution time as "call (entry) time - return (exit) time", which includes sub-calls?

example code:
```
def a():
    sleep(1 s)
    b()
def b():
    sleep(1 s)
    pass
a()
```
would lead to this trace:
```
[CALL] a() at 0 s
[CALL] b() at 1 s
[RETURN] b() at 2 s (duration 1s)
[RETURN] a() at 3 s (duration 3s)
```
actual total duration: 3 s
sum of function durations: 3 s + 1 s = 4 s

andrej (changes requested):
Looks mostly good. The only things I'd like to see addressed are removing the potentially unused code and making sure that in the total time, durations aren't double counted. I like the logger-based implementation. Simple and effective. :)

applications/llama_3.2_1b/analyze_profile.py
kurtis-b-1: Yeah, it was just using the duration from the log before. I removed the logging of the duration to reduce the size of the log file, so now this is used.

applications/llama_3.2_1b/analyze_profile.py
kurtis-b-1: Yes, it would. Basically, the total times logged here are cumulative times, including the sub-calls. This makes it so that the `generate()` function time matches the total time reported for inference for example. How I'm deciding for performance improvements currently is by finding how much certain functions take up the inference time.

applications/llama_3.2_1b/analyze_profile.py
kurtis-b-1: Removed comments

applications/llama_3.2_1b/analyze_profile.py
kurtis-b-1: Agreed, fixed this and now it's getting used.

applications/llama_3.2_1b/analyze_profile.py
kurtis-b-1: Agreed, these were also leftovers.

applications/llama_3.2_1b/inference.py
kurtis-b-1: I moved calculating the duration into `analyze_profile.py`, so this is no longer needed.

kurtis-b-1:
Thanks for looking through this! I was being a little lazy and didn't do code cleanup since everything worked...sorry about that. I did some cleanup and also made it so that only the start/end timestamps of function calls/returns are logged. Calculating the duration of functions have been moved inside of the analysis script.

applications/llama_3.2_1b/analyze_profile.py
andrej: I have no issue with using the total function time for each function to identify bottlenecks, that makes a lot of sense. What I don't understand is what summing them up and printing them as "total execution time" at the end is supposed to represent. Summing up these numbers double-/triple-/quadruple-... counts execution times and you can make the number arbitrarily big by wrapping functions in more functions.

kurtis-b-1: I'm not quite following. I guess the distinction you want to make is between cumulative total times with subcalls included and with subcalls excluded? Currently what's being logged is the former.

kurtis-b-1: Note for documentation: we discussed it and I now understand it's about the total sum across all execution times. I agree this isn't useful and it's been removed.

applications/llama_3.2_1b/analyze_profile.py
kurtis-b-1: That's true. Honestly none of that logic was needed, so I removed it.

andrej (approved):

PR #24, opened 2025-11-18
Fix CI PR comments permission issue and handle cancellations gracefully

andrej:
## Changed
- Replace the GitHub script implementation of "download artifact" with the dedicated "actions/download-artifact" action. The current implementation receives a 403 Forbidden error when trying to download the artifact; this new simpler implementation (which uses the GitHub token for authentication) hopefully fixes that. Unfortunately we will only know whether this was fixed once it is merged in `devel`.
- The workflow that adds comments to PRs with the test/benchmark results will fail when the workflow that triggered it is cancelled (because there are no artifacts). This change handles such cancellations gracefully.
- Also handle gracefully if the workflow that triggered this was not from a PR (e.g. nightly runs or pushes to main).

## PR Merge Checklist

### DIFF
--- a/.github/workflows/pr-comment.yml
+++ b/.github/workflows/pr-comment.yml
@@ -23,44 +23,28 @@ jobs:
     steps:
       - name: Download artifact
         id: download
-        uses: actions/github-script@v5
+        continue-on-error: true
+        uses: actions/download-artifact@v5
         with:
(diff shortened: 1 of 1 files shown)


hunhoffe (approved):
Hoping it works!

PR #25, opened 2025-11-18
MHA: Support for any Sequence Length

kurtis-b-1:
This PR adds support for arbitrary sequence length for the MHA via padding. We pad to the block size (64) times the number of pipeline used. The work in this PR was done by @Victor-Jung.

### Added
- Add kernel support for padding mask
- C++ testbench and golden model generation scripts are now aware of the number of pipeline used to adapt padding accordingly
- C++ testbench takes into account padding for value testing
- Golden model pads the I/O accordingly.

Note: A smarter implementation of the handling of the Q blocks in the case of multiple pipelines would allow to pad only to the block size.

## PR Merge Checklist


jgmelber (approved):

andrej:
image

Pretty crazy speedups. Very cool, unless there's a bug? Is this expected @Victor-Jung ?

@kurtis-b-1 Would you happen to have tested if this works in the llama model with other prompt lengths?

kurtis-b-1:
> @kurtis-b-1 Would you happen to have tested if this works in the llama model with other prompt lengths?

Tried it just now with smaller `--prompt_len 128` and `--prompt_len 64`, but it resulted in NaNs in the output. If I disable fused MHA then everything works. Looked into it some, and it seems like to support lower sequence lengths in the MHA operator (`aie_mha.py`), some padding needs to be added there. Should I keep looking into it?

andrej:
If you don't mind, yes please. Victor probably will have some insights, so maybe wait for him to respond.

Victor-Jung:
> image
> Pretty crazy speedups. Very cool, unless there's a bug? Is this expected @Victor-Jung ?
> 
> @kurtis-b-1 Would you happen to have tested if this works in the llama model with other prompt lengths?

The number of GFLOPS is correct, most likely the MHA was previously misconfigured (using 1 pipeline or something)

Victor-Jung:
> If you don't mind, yes please. Victor probably will have some insights, so maybe wait for him to respond.

I thought there was maybe an issue when using 8 pipelines for short sequence lengths so I tested the head of `pr/reminder-tile-mha` with sequence length of 64 and 128 and it works. I think something wrong happened in Curt's integration. I suggest to try using 1 pipeline, if it works, then we use 1 pipeline for short sequence lengths.

kurtis-b-1:
I had to add some padding logic into the AIE MHA operator. It should be possible to run inference with lower sequence lengths (tested `prompt_len` set to 64, 128, 256) now, though I believe the `prompt_len` value still needs to be a multiple of 64. 
Also I had to make a minor change to GEMM--basically there's an optimization to use as many BDs for the transfer blocks in the runtime sequence, which becomes a `repeat` parameter in the tiler. But if `M==m*n_aie_rows`, then there's no need to repeat since the NPU generates the full row outputs. So I added a `min()` to decide the value for that particular `repeat` parameter.

PR #27, opened 2025-11-19
Pin milr-aie version to a release and create action to keep it updated.

hunhoffe:
This PR sets the version of mlir-aie to the most recent stable release in the ```requirements.txt``` file. This effectively pins the version.

This PR also creates a github action that runs as a chron job every night at midnight. It then attempts to update the version of mlir-aie if there is a new release; in this case, it will create a PR to stage the update to the new release, allowing humans to review the change before it is merged in. The action can also be triggered manually.

Unfortunately, I am somewhat testing in prod (or rather, the devel branch) because it needs to be merged before this action can be fully tested.

## Added
- Action to create PR for mlir-aie releases; will run overnight daily

## Changed
- mlir-aie version from latest to latest release

## Removed
- N/A

## PR Merge Checklist


.github/workflows/update-mlir-aie.yml
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Check for new mlir-aie release
+
+on:
+  workflow_dispatch: # Allow manual triggering
+  schedule:
+    - cron: '0 0 * * *' # Run every day at midnight
```
andrej: This is midnight UTC, which is 5/6 PM Mountain Time. Perhaps it might be better to move a little later to when no one might be working late?

.github/workflows/update-mlir-aie.yml
```diff
@@ -0,0 +1,73 @@
...
+        id: current_version
+        run: |
+          version=$(grep 'mlir_aie==' requirements.txt | cut -d'=' -f3)
+          echo "Current version: $version"
+          echo "version=$version" >> $GITHUB_OUTPUT
+
+      - name: Get latest mlir-aie release
+        id: latest_release
+        run: |
+          latest_tag_with_v=$(curl -s "https://api.github.com/repos/Xilinx/mlir-aie/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
```
andrej:
```suggestion
          latest_tag_with_v=$(curl -s "https://api.github.com/repos/Xilinx/mlir-aie/releases/latest" | jq '.tag_name')
```

Probably more resilient JSON parsing than grep+sed

.github/workflows/update-mlir-aie.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Compare versions and create PR
+        if: steps.current_version.outputs.version != steps.latest_release.outputs.tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "New version found: ${{ steps.latest_release.outputs.tag }}. Current version is ${{ steps.current_version.outputs.version }}."
+          
+          # Configure git
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
```
andrej:
```suggestion
          git config --global user.name 'github-actions[bot]'
          git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
```

[Seems to be](https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-using-the-built-in-token) the correct e-mail for the actions bot.

andrej (approved):
Looks good to me.

Could we also do this for un-tagged wheels, i.e. test the latest commit nightly? Or is that something you'd rather do in MLIR-AIE?

.github/workflows/update-mlir-aie.yml
hunhoffe: Good call, thanks!

.github/workflows/update-mlir-aie.yml
hunhoffe: Thanks again!

.github/workflows/update-mlir-aie.yml
hunhoffe: I'm more likely to be working at 6 AM rather than midnight, but I think 3 AM might be a nice middle ground!

hunhoffe:
@andrej I think it's best to do this by release just so there is less noise... @jgmelber, thoughts?

Also, I realized that the wheels that relate to the releases aren't tagged with the release (they are tagged normally, with the commit) so I should probably address that in mlir-aie. The practical result of this is that pip won't let you check for a version of the wheels with a release tag, which is kind of a bummer.

hunhoffe:
It is failing now because changes from mlir-aie have already been integrated that are more recent than the previous release. This is a good sign things are working as expected! I will cut a new release and try again tomorrow

hunhoffe:
Wheels with version matching the release tags are in: https://github.com/Xilinx/mlir-aie/pull/2722

I'll pause this PR until I can create a properly versioned release.

hunhoffe:
@andrej I thought more about your comment and I think a good path forward might be:
* Keep this PR as-is, so only releases are tested by default
* Figure out how to create an action that can be triggered manually that can test a specific wheels version. This is useful for chasing down the root cause of any regressions that happen; however, only doing this on demand reduces the repo "noise" (e.g., accumulation of data, etc.) I would like to do that in a separate PR, if that's ok?

hunhoffe:
Using a new, stable release of mlir-aie with the version set correctly. I think it's ready to merge!

PR #31, opened 2025-11-21
Fix AIE Ops in the GQA Block

kurtis-b-1:
The attention projections are GEMM operations that are still passing in the weights during the forward pass when offloaded to the NPU, but those weights can be allocated before the forward pass. Also, when reshaping the output of the AIE-offloaded MHA operation, the wrong dimensions are used, which affects prefill staged output.

## Changed
- Set the weight values for the attention projection operations before the forward pass
- Use the correct dimensions for transposing the AIE MHA output

## PR Merge Checklist


jgmelber (approved):

andrej (approved):
🎆 🍾

PR #37, opened 2025-11-26
Unify Operators using Python Interface, Remove CMake

andrej:
- Drop the CMake-based structure of the existing operators in favor of the Python interface that we devised for the llama application, i.e. move from `CMakeLists.txt` to wrapping all operators in a Python AIE base operator.
- Port all `test.cpp`s to `test.py`s, aiming to factor out shared/previously duplicated code.
- Restructure into an `operators` directory, where each operator has a subdirectory with a `design.py` (IRON code), `op.py` (operator wrapper/Python interface), `reference.py` and `test.py`.
- <s>Add a "dry run" facility to the Python interface that allows printing the compilation commands for artifact generation. I added this primarily because it was useful to verify against the CMake compilation commands.</s> Removed, commits were 0fa2fca, 350dd5f, 2250e79 and 69994ff.
- Small cleanups as I touched operators, e.g. disambiguate AIE (hardware) columns and matrix columns with clearer variable names.
- Add test discovery script that works without CMake to run the `test.py`s in CI.

Remaining to do:
- WIP: Add support for data types other than `bf16` in the Python operator interface -- needed for the `dequant` kernel.
- Port remaining two operators (memcpy, dequant).
- Remove all remnants of CMake.
- Make the CI call the new Python test discovery script rather than CMake.


applications/llama_3.2_1b/src/compilation.py
```diff
@@ -198,40 +208,105 @@ def compile(self, artifacts):
...
+                if artifact.requires_context:
+                    python_cmd += "from aie.extras.context import mlir_mod_ctx; "
+                    python_cmd += "with mlir_mod_ctx() as ctx: "
+                python_cmd += f"mlir_code = {artifact.callback_fn}({', '.join(map(GenerateMLIRFromPythonCompilationRule._repr_for_codegen, artifact.callback_args))}, {', '.join(f'{k}={_repr_for_codegen(v)}' for k, v in artifact.callback_kwargs.items())}); "
+                if artifact.requires_context:
+                    python_cmd += "print(str(ctx.module))"
+                else:
+                    python_cmd += "print(str(mlir_code))"
+                self.dry_run.append(f"python3 -c '{python_cmd}' > {artifact.path}")
+                new_artifact.fake_available = True
```
hunhoffe: This path stuff is gnarly. I'm not saying it needs to be fixed in this PR (since this PR is already large) but I'd be interested to know more about why this type of loading on the fly is necessary. I wonder if you could add the operator paths in the venv setup....?

applications/llama_3.2_1b/src/compilation.py
```diff
@@ -198,40 +208,105 @@ def compile(self, artifacts):
...
+        obj_module = type(obj).__module__
+        obj_class = type(obj).__name__
+
+        # Check for known device class patterns (NPU1, NPU2, XCVC1902, etc.)
+        # These are imported from aie.iron.device but have __module__ == 'abc'
+        if obj_module == "abc" and (
+            obj_class.startswith("NPU") or obj_class.startswith("XCVC")
+        ):
+            # For device classes, generate instantiation code
+            return f"{obj_class}()"
```
hunhoffe: I think you can use ```issubclass``` to check if it's a ```Device```, if I am understanding this code properly.

e.g.,
```python
>>> from aie.iron.device import NPU1, NPU2
>>> from aie.iron.device import Device
>>> issubclass(NPU1, Device)
True
```

operators/common/discover_tests.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python3
```
hunhoffe, line 1: this could be a good intermediary step, but is there a reason we can't use something like pytest to find tests within a directory? This seems a little like reinventing the wheel when there are mature python testing libraries.

Another option could be to find classes with tests during module load time in the operator top level __init__ file (e.g., do introspection to find which of the loaded things are testable).

I'm fine with this for now -- just brainstorming :)

operators/common/utils.py
```diff
@@ -0,0 +1,323 @@
...
+        top_k (int, optional): The number of top logits to consider for sampling. If None, all logits are used. Default is None.
+        eos_id (int, optional): The end-of-sequence token ID. If specified, generation will stop when this token is produced. Default is None.
+
+    Returns:
+        torch.Tensor: The updated sequence of token indices after generation (shape: (batch_size, new_sequence_length)).
+    """
+    # For-loop is the same as before: Get logits, and only focus on last time step
+    finished_prefill = False
+
+    print(f"Starting prefill inference...")
```
hunhoffe: Use logger instead of print, maybe?

operators/gemm/design.py
```diff
@@ -511,16 +522,6 @@ def core_fn(in_a, in_b, out_c, zero, matmul, convert_copy, my_rtp, barrier):
             tile_group_col_major=True,  # Send all tiles in column before moving on to next column
             prune_step=False,
         )
```
hunhoffe, line 524: Are the changes purposeful (or because of need to rebase)?

operators/gemv/test.py
```diff
@@ -0,0 +1,94 @@
...
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent))
+
+from operators.gemv.op import AIEGEMV
+from operators.gemv.reference import generate_golden_reference
+from operators.common.test_utils import run_test
+
+
+regular_test_cases = [
```
hunhoffe, line 16: This could (in a later PR probably) be addressed with something like a ```pytest.parametrize``` with ```M```, ```K```, etc. parameters.

operators/relu/op.py
```diff
@@ -0,0 +1,116 @@
...
+        assert total_shimdma_channels <= 16, "Conservative ShimDMA limit"
+
+        self.xclbin_artifact = None
+        self.insts_artifact = None
+
+        AIEOperatorBase.__init__(self)
+
+    def set_up_artifacts(self):
+        operator_dir = Path(__file__).parent
+        file_name_base = f"relu_{self.num_aie_columns}c_{self.num_channels}ch_{self.size}_{self.tile_size}t"
```
hunhoffe, line 42: I wonder if both ```file_name_base``` and ```mlir_artifact``` could have a default impl in the operator class that simplifies this setup for most operators (could. be overridden if needed).

I also wonder if the set_up_artifacts should be part of initialization, but that's probably a question for another day.

hunhoffe (approved):
I think this is a great step forward! 

I'd recommend subsequent PRs to:
* Simplify test setup/config, probably using a library such as pytest
* Simplify the python import/path setup in this repo

However, this PR is already large/chaotic, so I'd review my comments but push these tasks to future work.

operators/relu/op.py
ypapadop-amd: It'd be great if there was some customization capability to be able to change the name / directory of object files and MLIR files.

operators/common/compilation.py
```diff
@@ -0,0 +1,630 @@
...
+
+                if artifact.requires_context:
+                    python_cmd += "from aie.extras.context import mlir_mod_ctx; "
+                    python_cmd += "with mlir_mod_ctx() as ctx: "
+                python_cmd += f"mlir_code = {artifact.callback_fn}({', '.join(map(GenerateMLIRFromPythonCompilationRule._repr_for_codegen, artifact.callback_args))}, {', '.join(f'{k}={_repr_for_codegen(v)}' for k, v in artifact.callback_kwargs.items())}); "
+                if artifact.requires_context:
+                    python_cmd += "print(str(ctx.module))"
+                else:
+                    python_cmd += "print(str(mlir_code))"
+                self.dry_run.append(f"python3 -c '{python_cmd}' > {artifact.path}")
```
ypapadop-amd, line 280: You can get the python executable with `sys.executable`

operators/common/compilation.py
```diff
@@ -0,0 +1,630 @@
...
+            return f"{obj_class}()"
+
+        # Default to repr() for other types
+        return repr(obj)
+
+
+class AieccCompilationRule(CompilationRule):
+    def __init__(self, build_dir, peano_dir, mlir_aie_dir, *args, **kwargs):
+        self.build_dir = build_dir
+        self.aiecc_path = Path(mlir_aie_dir) / "bin" / "aiecc.py"
```
ypapadop-amd, line 314: You can probably reuse some of the logic in https://github.com/Xilinx/mlir-aie/blob/9db8a590303c59bd88be58f0596f15afb4656af7/python/iron/compile/compile.py#L79 to avoid an extra shell.

operators/common/compilation.py
```diff
@@ -0,0 +1,630 @@
...
+            isinstance(artifact, KernelObjectArtifact)
+            and all(
+                isinstance(dependency, SourceArtifact) and dependency.is_available()
+                for dependency in artifact.depends
+            )
+            for artifact in artifacts
+        )
+
+    def compile(self, artifacts):
+        clang_path = Path(self.peano_dir) / "bin" / "clang++"
```
ypapadop-amd, line 448: Peano compiler, linker, etc. are exposed in https://github.com/Xilinx/mlir-aie/blob/9db8a590303c59bd88be58f0596f15afb4656af7/python/utils/config.py#L13

ypapadop-amd:
It's kind of big to review exhaustively, but I think it's a good step towards a reusable IRON framework. We have a number of functions for compilation, linking etc. in MLIR-AIE that can be reused.

applications/llama_3.2_1b/src/compilation.py
andrej: Agreed that having the operators in the path would be the correct solution. This is a bit of a remainder from when the operators were in the `examples` directory and we were running llama from a completely different directory. As for the commands being generated for the dry-run, those are intended to be able to be ran stand-alone in a terminal to produce the same artifacts. Created issue #42  for this.

applications/llama_3.2_1b/src/compilation.py
andrej: Much better indeed. Maybe I'll rip this dry run stuff out altogether...

operators/common/discover_tests.py
andrej: Claude whipped this together with me so these tests could fit into the existing testing framework that is a weird handmade solution of CMake+Python. Once it's all migrated to Python, I agree that we should move to something more established. Created issue #42 for this which I can tackle after merging this.

operators/common/utils.py
andrej: Actually these are mostly llama-related. Thanks for drawing my attention to this file. I'll move it back to llama where it belongs.

operators/gemm/design.py
andrej: These are Curt's changes that re-enabled C-column major GEMM. Don't know why the diff is showing up weird, but I hand-verified them to match. There also shouldn't be conflicts with these files since they contain Curt's changes, so I messed something up git-wise here.

operators/gemv/test.py
andrej: Agreed. This is ugly and for now just mirrors what was in the CMake directly and I didn't like doing it but as you already pointed out, the PR is already getting large.

andrej:
Thanks for the comments @ypapadop-amd. I agree it's large. The goal was to limit the scope of this PR to only removing the duplication of build scripts for the operators. Instead of having the CMake build scripts under `example` and the Python build scripts in `application/llama`, this moves everything to use the Python-based solution.

The `compilation.py` etc. were mostly just moved from where they currently sit in `applications/llama` to `operators/common`. 

Erika and I are planning for this to be a multi-step refactor. Reusing existing code from elsewhere (more code deduplication) and coming to a clean, reusable operator interface are the next steps. Perhaps addressing things like operator artifact naming, and reusing existing IRON and JIT code should be part of those next steps.

applications/llama_3.2_1b/src/compilation.py
andrej: dry run stuff has been removed

applications/llama_3.2_1b/src/compilation.py
andrej: dry run stuff has been removed

PR #41, opened 2025-12-01
Update gemm tiling dim (n) and move function to appropriate spot

kurtis-b-1:
For some reason tiling dim `n` for GEMM in CMake was `32` instead of `64`--updated it to be `64`.

## Changed
- Gemm tiling dimension `n` increased to `64`
- The `add_gemm_archive` function was in an incorrect spot, so moved it with the other functions

## PR Merge Checklist


kurtis-b-1:
@andrej Not sure if this is needed since you're removing the CMake build system, but created this PR just to point to a small issue in the tests for GEMM.

andrej:
The default `n` dimension is already changed to 64 in #37 :)

PR #44, opened 2025-12-04
Switch to pytest

andrej:
## Added
- pytest test harness.
- AIE Context class which manages runtime and compile artifacts. Before, there could only be one instance per Python execution, which broke pytest which runs tests one-after-another in the same execution.

## Removed
- Old hand-built test harness.

## PR Merge Checklist


operators/dequant/test.py
```diff
@@ -84,13 +83,20 @@ def main():
...
+    Latency=r"Latency \(us\): (?P<value>[\d\.]+)",
+    Bandwidth=r"Effective Bandwidth: (?P<value>[\d\.e\+-]+) GB/s",
+)
+@pytest.mark.extensive
+@pytest.mark.parametrize(
+    "input_length,num_aie_columns,num_channels,tile_size,group_size",
+    extensive_params,
+    ids=extensive_names,
+)
+def test_dequant_extensive(
```
hunhoffe: I have not tried this myself, but rather than having two functions, can you use a combination of mark + parametrize to make this work with just one function definition?
```python
@pytest.mark.parametrize("input_value", [
    10,
    pytest.param(12, marks=pytest.mark.suiteB),
    pytest.param(13, marks=pytest.mark.suiteC),
])
```

hunhoffe (approved):
Overall, a good step forward (and very speedy!)

I do wonder if a little further simplification is possible. One is whether you can make the parameterization of the suites more streamlined (I wrote a comment about this). The other is whether the context is really needed... would a session-scoped fixture serve the same purpose?

e.g., something like,
```python
@pytest.fixture(scope="session")
def setup_once_per_suite():
    """
    This fixture will run only once for the entire pytest session.
    """
    print("\nSetting up resource once per suite...")
    resource = "Shared resource"
    yield resource
    print("Tearing down resource once per suite.")
```

I have mixed feelings about this. I think generally the context is somewhat of a good idea, so maybe we want to keep it even if pytest may offer a workaround. On the other hand, simplicity is good.

andrej:
Thanks for the review.

> The other is whether the context is really needed... would a session-scoped fixture serve the same purpose?

Having the "context" (before it was called that) be session-scoped is exactly what was the issue before (all AIEBaseOperators shared the same state of registered operators, no explicit context means all had the same context). If we test operator A, then operator B, within the same Python session, we need to "reset" the XRT runtime buffers etc. that we have set up, because B will use different buffers than A. So that's what having one fixture per test is trying to fix, rather than per session, and why I added the context.

Open to suggestions on how to simplify this further.

Will look into your other comment. 👍

hunhoffe:
Ah, so session scoped is bad, makes sense... what about function scoped, e.g., ```scope="function"``` in pytest fixtures?

andrej:
I think that's what I'm [already doing](https://github.com/andrej/IRON/blob/58b054423eb4ab4fa8d22e0f81662d236c9c5845/conftest.py#L17)?

hunhoffe:
I think your confusion is accurate, because I think I am missing something in my understanding. To check my knowledge: you are allocating a bunch of buffers, compiling all the operator kernels, and flushing the entire cached kernel device every time the context is refreshed which happens at function scope for every single pytest? 

I had thought all you'd need to do is flush the cache for the device handler, and then also maybe pre-allocate a small number of buffers and compile the operator as part of each test. Maybe this is not true?

hunhoffe:
To be clear, I only really care about these details to ensure the test infrastructure has the following properties:
* The tests don't take longer to run than they need to
* We have a property where if you break one operator, only the tests directly using/targetting that operator break (modularity).

andrej:
I think I'm also confused. What do you mean by "cached kernel device" and "kernel is refreshed"?

What the current implementation does: Collect a new set of artifacts for each operator and set up a new set of buffers for each operator. A set of artifacts and set of buffers for a set of operators you plan to call is a "context." For these tests, each context should only contain a single operator. Concrete example: 1. test foo, 2. then test bar. Results in: context 1: artifacts "foo.o, foo.xclbin, foo.mlir", buffers "foo_input", "foo_output". context 2: artifacts "bar.o, bar.xclbin, bar.mlir", buffers "bar_input", "bar_output"...

What the previous implementation did (single implicit context): Gather all artifacts and buffers across all operators into one context. Meaning you'd end up with artifact list "foo.o, foo.xclbin, foo.mlir, bar.o, bar.xclbin, bar.mlir" and buffers "foo_input, foo_output, bar_input, bar_output".

Potential issues with this: name collisions of artifacts (should be solved more robustly but isn't currently), allocating too many buffers at once (since we're allocating all buffers for all operators of the context ahead of time, even if we don't call all of those operators) resulting in an out of memory condition. 

Concrete issue that I actually faced: You can currently only call "prepare_runtime" once since it goes through and allocates all buffers. If you try to do it a second time, it'll try to reallocate the same buffers and I think that's where things failed for me. Since I want to run operator test A first, then operator test B, I need to set up a runtime for test A first, but then I can't test operator B since the runtime is already set up. The way around this would be to first set up both operators, then test both operators, (i.e. `setup A, setup B, test A, test B` rather than `setup A, test A, setup B, test B`) but to me, it seemed like an unnecessary/weird restriction that we can't just have multiple contexts.

kurtis-b-1 (approved):

hunhoffe:
I was talking about the kernel being loaded already into the cache managed by the device_manager.

But if you have a test that only tests one operator, why do you gather everything for all operators? If that is indeed what is happening, then you don't have modularity in the tests between operators?

Sorry for my confusion! It seems like what you are doing is probably reasonable, I just don't understand it.

andrej:
> * The tests don't take longer to run than they need to

This pytest implementation should be the same or slightly better runtime-wise as was we currently have in the repository. Since each test currently runs as a standalone Python script (called using `subprocess`), each test currently has its own completely isolated context. The change in this PR doesn't create suprocesses but instead runs everything in one Python file, so the singleton context got crowded with every test's artifacts and buffers, which is what the AIEContext addresses.

> * We have a property where if you break one operator, only the tests directly using/targetting that operator break (modularity).

The added AIEContext should help with this. We lost some of this isolation by not running separate Python scripts, so adding the AIEContext should restore some of that isolation.

kurtis-b-1:
> I had thought all you'd need to do is flush the cache for the device handler, and then also maybe pre-allocate a small number of buffers and compile the operator as part of each test. 
>
> What the current implementation does: Collect a new set of artifacts for each operator and set up a new set of buffers for each operator. A set of artifacts and set of buffers for a set of operators you plan to call is a "context." For these tests, each context should only contain a single operator. Concrete example: 1. test foo, 2. then test bar. Results in: context 1: artifacts "foo.o, foo.xclbin, foo.mlir", buffers "foo_input", "foo_output". context 2: artifacts "bar.o, bar.xclbin, bar.mlir", buffers "bar_input", "bar_output"...

Hi guys, I'm also trying to understand...I think these are the same idea?

andrej:
> But if you have a test that only tests one operator, why do you gather everything for all operators? If that is indeed what is happening, then you don't have modularity in the tests between operators?

That is exactly what the `AIEContext` is trying to solve.

What you describe is the current state of the repository. Look [here](https://github.com/amd/IRON/blob/130b6ea70fafd6fa09d3d22acdde311f0e509784/operators/common/aie_base.py#L22) -- the current implementation uses a global variable to keep track of all registered operators and compiles and sets up the runtime for all of them. 
The intent behind this is that multiple operators can share artifacts (e.g., shared kernel objects, cutting down compilation time) and  XRT buffers (cutting down memory usage).

However, this broke the isolation of tests. So rather than collecting all registered operators in one global state, you now get to specify where to collect them into (i.e., the AIEContext).

Again, to be clear, these tests add and test just a single top-level operator in each context. (Tests with combined operators like SwiGLU do create multiple sub-operators, but that is how they are meant to work.)

andrej:
> Sorry for my confusion! It seems like what you are doing is probably reasonable, I just don't understand it.

Don't apologize for scrutinizing the code and helping make it better!

andrej:
> Hi guys, I'm also trying to understand...I think these are the same idea?

Yeah I think you might be right. I think the code does what Erika says it should do, it's just probably unnecessarily confusing.

hunhoffe:
Yes! I think it get it now. I was misunderstanding how the operation registration works in this model. I thought the operators were automatically registered, but if they are registered per test per context on demand, this makes sense I believe!

andrej:
Open to any suggestions to make this clearer. A lot of this is just a chain of hacks stringed together.

And to confirm your understanding yes, test registration happens in the constructor of the operator. (Think of each parametrization of an operator as a separate operator, so we can only register once it's constructed with concrete parameters.) Since that constructor is called in the test itself, and each test gets its own context via the fixture (`aie_context` argument) that it passes to the constructor, each test is registered in its own context.

hunhoffe:
I think this is plenty good for now, it was just unclear to me why the context needed to be so complex. I think I understand now that it encompasses a lot of what would otherwise be boilerplate. Fine-tuning can happen later as needed, IMO.

andrej:
A lot of the context runtime set up logic has just moved. That complexity was previously in AIEBaseOperator. The reason the buffer allocation is complex is because it (a)  tries to find a small set of buffers to allocate and share between operators, while (b) also allowing the "static data" optimization (i.e., buffers that cannot be shared between operators because their data is weights that should not be changed). I picture a lot of this will go away once the IRON Tensor makes its way here (I think the "sharing buffers" optimization might get dropped).

Bigger picture comment: My vision is  that operators are as declarative (as opposed to imperative) as possible. The idea in this version of the operator is that `get_artifacts` and `set_up_runtime` only "set some variables" to _describe_ the artifacts and the runtime buffers needed (not actually set them up, maybe I gave the functions bad names and `artifact_desc()` and `runtime_desc()` would be better-suited).

My hope is that in that way, it should be easier to swap out backends: e.g., from the buffer description (buffer names + sizes + runlist), you could both generate a C++ host code to run it or use PyXRT or HSA. All consume the same runtime description (a static list of buffer names, sizes, runlist sequence). Same for the artifacts: You can use our current Python-based compilation process, or use the declaration of artifacts to instead "export" a Makefile or even CMakeLists.txt (not yet implemented, but could be done).

The declarative nature of each operator makes the operator description slimmer. It means that the "backend implementation" of how that declaration gets turned into imperative instructions (e.g. "allocate this buffer using XRT, call this kernel, ...") gets pushed elsewhere, in this case the context. Otherwise, we'd have this logic embedded in each operator (and, in turn, it might be harder to swap out backends).

hunhoffe:
That makes a lot of sense to me! Thank you for elaborating on your vision of how the context fits into the larger picture -- it did help clarify some things for me!

Issue #43, opened 2025-12-04
XDNA driver/XRT regression causing `ERT_CMD_STATE_TIMEOUT` in runlist execution


alxchk:
```
(iron-venv) llm@dagger ~/tools/IRON/applications/llama_3.2_1b $ xrt-smi validate
WARNING: User doesn't have admin permissions to set performance mode. Running validate in Default mode
Validate Device           : [0000:c5:00.1]
    Platform              : NPU Strix
    Power Mode            : Default
-------------------------------------------------------------------------------
Test 1 [0000:c5:00.1]     : gemm 
    Details               : TOPS: 51.0                                          
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Test 2 [0000:c5:00.1]     : latency 
    Details               : Average latency: 50.0 us                            
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Test 3 [0000:c5:00.1]     : throughput 
    Details               : Average throughput: 87380.0 op/s                    
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Validation completed. Please run the command '--verbose' option for more details
(iron-venv) llm@dagger ~/tools/IRON/applications/llama_3.2_1b $ 
```

```
        /$$$$$$ /$$$$$$$   /$$$$$$  /$$   /$$
       |_  $$_/| $$__  $$ /$$__  $$| $$$ | $$
         | $$  | $$  \ $$| $$  \ $$| $$$$| $$
         | $$  | $$$$$$$/| $$  | $$| $$ $$ $$
         | $$  | $$__  $$| $$  | $$| $$  $$$$
         | $$  | $$  \ $$| $$  | $$| $$\  $$$
        /$$$$$$| $$  | $$|  $$$$$$/| $$ \  $$
       |______/|__/  |__/ \______/ |__/  \__/
                                            
                                            
 /$$       /$$        /$$$$$$  /$$      /$$  /$$$$$$ 
| $$      | $$       /$$__  $$| $$$    /$$$ /$$__  $$
| $$      | $$      | $$  \ $$| $$$$  /$$$$| $$  \ $$
| $$      | $$      | $$$$$$$$| $$ $$/$$ $$| $$$$$$$$
| $$      | $$      | $$__  $$| $$  $$$| $$| $$__  $$
| $$      | $$      | $$  | $$| $$\  $ | $$| $$  | $$
[...]

PR #50, opened 2025-12-05
Fix padding in SwiGLU operator; fix call to softmax in llama

andrej:
A couple small stability fixes in response to #43.

## Changed
- Runlist-based SwiGLU is now padding-aware. Previously, its sub-operators would silently add padding but SwiGLU would not account for this in its allocated buffer sizes, leading to memory corruption by XRT.
- The softmax kernel constructor was slightly changed in a recent PR and the change hadn't percolated to the callsite in llama. Thsi fixes that, so that `regular_mha: true` works again.

## PR Merge Checklist

### DIFF
--- a/operators/softmax/op.py
+++ b/operators/softmax/op.py
@@ -21,16 +21,14 @@ from operators.common import (
 class AIESoftmax(AIEOperatorBase):
 
     def __init__(
-        self,
-        rows: int,
-        cols: int,
-        num_aie_columns=1,
-        num_channels=1,
-        tile_size=None,
-        context=None,
+        self, 
+        rows: int, 
+        cols: int, 
+        num_aie_columns=1, 
+        num_channels=1, 
+        context=None
     ):
         self.size = rows * cols
-        self.tile_size = tile_size if tile_size is not None else cols
         self.rows = rows
         self.cols = cols
 
@@ -46,7 +44,7 @@ class AIESoftmax(AIEOperatorBase):
     def set_up_artifacts(self):
         # Compilation artifacts
         operator_dir = Path(__file__).parent
-        file_name_base = f"softmax_{self.num_columns}c_{self.num_channels}ch_{self.size}_{self.tile_size}t"
+        file_name_base = f"softmax_{self.num_columns}c_{self.num_channels}ch_{self.size}_{self.cols}t"
 
         mlir_artifact = PythonGeneratedMLIRArtifact.new(
             f"{file_name_base}.mlir",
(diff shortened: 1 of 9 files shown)


operators/swiglu_prefill/op.py
```diff
@@ -211,9 +237,31 @@ def _execute_aie_operation(self, x):
         # Flatten inputs for AIE processing
         x_flat = x.view(-1)
 
+        # Verify input size matches expected dimensions
+        expected_size = batch * self.seq_len * self.embedding_dim
+        assert x_flat.shape[0] == expected_size
```
AngryLoki, line 236: This fails when prompt is short:
```
python applications/llama_3.2_1b/inference.py ~/models/model.safetensors ~/models/tokenizer.model --prompt 'Hello!'

...
  File "/src/IRON/operators/swiglu_prefill/op.py", line 226, in forward
    out = self._execute_aie_operation(x_flat)
  File "/src/IRON/operators/swiglu_prefill/op.py", line 242, in _execute_aie_operation
    assert x_flat.shape[0] == expected_size
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Maybe the check should look like this?
```python
        # Verify input size does not exceed expected dimensions
        max_size = batch * self.seq_len * self.embedding_dim
        actual_size = x_flat.shape[0]
        assert actual_size <= max_size, f"Input size {actual_size} exceeds max {max_size}"
```

operators/softmax/op.py
```diff
@@ -20,11 +20,8 @@
 
 class AIESoftmax(AIEOperatorBase):
 
-    def __init__(
-        self, rows: int, cols: int, num_aie_columns=1, num_channels=1, tile_size=None
-    ):
+    def __init__(self, rows: int, cols: int, num_aie_columns=1, num_channels=1):
         self.size = rows * cols
```
kurtis-b-1, line 26: Are the normalization operators structured like this now too?

kurtis-b-1 (approved):
Had a small comment but looks good. Nice work on finding the memory corruption!

operators/softmax/op.py
andrej: I think only softmax is. The others still would need to be updated.

operators/swiglu_prefill/op.py
andrej: Thanks for pointing to this. It should be fixed for sequence lengths > 1. We'll tackle sequence length == 1 later.

PR #62, opened 2026-01-06
Support other tile sizes in GEMM `test.py`

andrej:
## Added
-

## Changed
- Uses the input `m, k, n` parameters to the GEMM test function to actually use those tile sizes, so if we want to add tests for other tile sizes in the future, or for local testing, we can.
- 
## Removed
-

## PR Merge Checklist


hunhoffe (approved):
Thank you for looking into this!

andrej:
Closing in favor of #65 which contains this change and others.

PR #71, opened 2026-01-26
Llama reimplementation with fused transaction sequence

andrej:
## Added
- Python code that automatically generates a fused MLIR operator from an "invocation plan" of smaller operators using `aiex.configure` and `aiex.run`
- `llama_npu.py` implementation that makes use of that fusion. goal: all operators end-to-end run on NPU, no involvement of CPU whatsoever
- A couple of new operators needed to run everything on NPU: `strided_copy` and `repeat` -- not yet very optimized
- new simple baseline CPU llama implementation

## Changed
- streamlined some of the compilation infrastructure where I ran into limitations
- added masking option with run-time parameter to softmax

## Removed
- old llama implementation

## To-Do

- [ ] offload everything in prefill too -- there are still some CPU operations there 
- [ ] fuse prefill together too
- [ ] upgrade actions runner to use XRT after https://github.com/Xilinx/XRT/pull/9560, which this relies on
- [ ] upgrade this repository to use MLIR-AIE after https://github.com/Xilinx/mlir-aie/pull/2831, which this relies on
- [ ] figure out something about patching? currently, patching the fused transaction sequence for decode slows everything down a lot because we re-instantiate a new `xrt::elf` for every token. proper solution would be to use new scratchpad memory firmware feature to read run-time parameters from DDR
- [ ] re-test all the other operators after refactoring compilation infrastructure, might have broken

## PR Merge Checklist


ItsMick:
I've been doing Llama 3.2 1B decode on Strix Halo with IRON — hit the same xclbin switching problem (measured 2.7-5.8ms warm switch, killed every multi-xclbin approach). Ended up at 22 tok/s with a hybrid NPU/CPU split using int4 quantized supertile GEMV for Q/K/V/gate/up, CPU for the rest.

I have int4 GEMV operators that might be useful here since mv.cc is bf16 only. Fused on-tile dequant, all 32 tiles, group_size=32, double-buffered ObjectFIFO:
- 8192×2048: 441us, 21.4 GB/s aggregate
- 2048×2048: 193us, 12.2 GB/s
- Beats CPU at all tested sizes (0.07-0.61x)

Worth noting — I benchmarked batched int4 GEMV and compute/DMA are roughly balanced at batch=1 (~941us DMA vs ~793us compute per batch element). Dequant is expensive enough that batching only gets 1.3x at batch=4. The int4 kernel probably needs optimization to not become the bottleneck in a fused pipeline.

Happy to share code or help integrate. DIM_K as compile-time constant for the int4 path should be straightforward.

andrej:
This got merged as part of #76.

PR #86, opened 2026-03-24
Doc updates

ypapadop-amd:
Describe the intent of your PR here.

## Added
- AGENTS.md with instructions for development with AI agents.

## Changed
- Updated README.md

## Removed
- NA

## PR Merge Checklist


README.md
```diff
@@ -160,12 +170,114 @@ chmod +x .git/hooks/pre-push
...
 - Python formatting (black)
 - C++ formatting (clang-format)
 
 To bypass the hook if needed: `git push --no-verify`
 
+## Quick Start Example
+
+Here's a simple example using the AXPY operator (Y = aX + Y):
+
+```python
```
hunhoffe: I am not sure it's wise to include the code in the README, as it might be hard to maintain.

ypapadop-amd: I can remove it. An agent can keep it updated though.

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+
+```bash
+# 1. Source XRT (required for all operations)
+source /opt/xilinx/xrt/setup.sh
+
+# 2. Activate virtual environment
+source ironenv/bin/activate
+
+# 3. Install dependencies
+pip install -r requirements.txt
```
andrej: Would it make sense to separate out the environment _installation_ steps (which would also have to include `python3 -m venv ironenv`) from the environment _activation_ steps? Also I've had trouble with calling `pip` directly in some environments, for maximum compatibility I think `python3 -m pip` would help.

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+### Environment Variables
+
+- `IRON_EXAMPLE_WEIGHTS_DIR`: Path to model weights for applications (default: `/srv`)
+
+## Building and Testing
+
+### Run All Operators (non-extensive tests)
+
+```bash
+pytest iron/operators/ -m "not extensive"
```
andrej: For functional tests only I'd add an `--iterations 1` here.

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+
+- Automatically initializes `pyxrt.device(0)`
+- Caches contexts and kernels per xclbin path
+- Shared across all operators to avoid resource conflicts
+
+## Hardware Constraints
+
+### NPU Architecture Limits
+
+- **NPU1 (AIE2)**: 4 rows × 1-4 columns (AMD Ryzen AI Phoenix/Hawk Point)
```
andrej: 4 rows x 4 accessible columns

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+- **NPU2 (AIE2P)**: 4 rows × 8 columns (AMD Ryzen AI 300 Series "Strix Point", Ryzen AI 9 HX 370 "Strix Halo", Krackan)
+
+### Tile and Dimension Constraints
+
+Common operator parameters and their constraints:
+
+- `tile_size`: Typically 64, 128, 256, or 4096 (depends on operator and data type)
```
andrej, line 212: Not sure this is a hard requirement, perhaps this is more confusing than helpful?

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+1. Create directory in `iron/operators/<operator_name>/`
+2. Implement `op.py`:
+   - Subclass `MLIROperator`
+   - Implement `get_operator_name()`, `get_mlir_artifact()`, `get_kernel_artifacts()`, `get_arg_spec()`
+   - Add validation for dimension constraints (assert statements)
+   - Define tile sizes and column counts
+3. Implement `design.py`:
+   - Import from `aie.iron` (Program, Runtime, Worker, ObjectFifo, Kernel)
+   - Define function that builds MLIR-AIE design
+   - Use `range_()` for loops (not Python `range`)
```
andrej, line 251: Would a "why" be useful here for the agents? Make it clear that this is metaprogramming that will translate directly to MLIR constructs such as scf.for

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+   - Add `event0()` and `event1()` for performance profiling
+5. Implement `reference.py` with CPU reference
+6. Implement `test.py` with pytest tests
+   - Use `@pytest.mark.extensive` for slower/larger tests
+   - Use `verify_buffer()` from `iron.common.test_utils`
+7. Register operator in `iron/operators/__init__.py`
+
+## Operator Fusion
+
+IRON supports fusing multiple operators into a single xclbin for improved performance:
```
andrej:
```suggestion
IRON supports fusing multiple operators into a single ELF file. This improves performance enabling a single runtime dispatch for a chain of operators. This works only with the "full ELF" flow, which uses ELF files at runtime. The ELF files take the place of `xclbin`s.
```

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+    input_args={"in": size_in},
+    output_args={"out": size_out},
+    context=ctx
+)
+```
+
+Benefits of fusion:
+
+- Reduces host ↔ NPU data transfers
+- Shares compiled resources (kernels.a)
```
andrej:
```suggestion
```

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+- Eliminates intermediate buffer round-trips
```
andrej:
```suggestion
- Runs a chain of operators using a single host-side dispatch (one CPU/host interrupt after fusion vs. one interrupt per operator without fusion)
```

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+```python
+num_columns = 4
+chunk = total_elements // num_columns
+
+taps = [
+    TensorAccessPattern(
+        (1, total_elements),
+        chunk * i,  # offset for column i
+        [1, 1, 1, chunk],
+        [0, 0, 0, 1],
```
andrej:
```suggestion
        [1, 1, 1, chunk],  # sizes
        [0, 0, 0, 1],  # strides
```

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+```bash
+pytest iron/operators/
+```
+
+### Run Single Operator Test
+
+```bash
+pytest iron/operators/axpy/
+# OR directly execute the test.py
+./iron/operators/axpy/test.py
```
AngryLoki: This does not run the test (it just does basic syntax check). It was true in the past, but wrong since tests were rewritten with pytest. README.md has the same mistake - it was not updated.

AGENTS.md
```diff
@@ -0,0 +1,501 @@
...
+### Run Application Tests
+
+```bash
+pytest iron/applications/
+```
+
+### Run Specific Test Function
+
+```bash
+pytest iron/operators/gemm/test.py::test_gemm_basic
```
AngryLoki: Hallucinated, there is no `test_gemm_basic`

README.md
```diff
@@ -160,12 +170,68 @@ chmod +x .git/hooks/pre-push
...
+- Parallel execution across multiple AIE columns
+- Optimized data movement via ObjectFIFOs
+- Fused operations to minimize host-NPU transfers
+- Vectorized kernels using AIE intrinsics
+
+Run benchmarks:
+
+```bash
+# Run all operators with performance metrics
+pytest iron/operators/ -m "not extensive" -v
```
AngryLoki, line 221: Better just mention that any run of `pytest` stores metrics in `tests_latest.csv` file. Adding `-s` to `pytest` allows to output these metrics to stdout, but it is still easier to use `csv` (because with csv it calculates avg/stddev). `--iterations` can be increased to increase stability.

AGENTS.md
ypapadop-amd: I don't think I understand the difference between the two.

AGENTS.md
ypapadop-amd: I found that I need to give it something for `tile_size` in designs outside IRON. Are the values too restrictive or just wrong?

AGENTS.md
ypapadop-amd: I am not sure. These are instructions so that agents can understand the codebase. Understanding MLIR may be outside the scope of that (?)

AGENTS.md
ypapadop-amd: There was one in my local changes 😄 I think it preferred that since I was asking a lot of questions about GEMMs.

jgmelber (approved):

AGENTS.md
andrej: In the other architecture you state 4 rows x 8 columns whereas here its 4 rows x 1-4 columns. But all Phoenix and Hawk Point devices have 4 rows x 4 columns as far as I know. Accessible you can leave off I just added that because technically there's 5 columns in hardware but the driver dropped support for using the first a while ago.

AGENTS.md
andrej: They're not wrong, if it's helpful should be fine to leave them in.

AGENTS.md
ypapadop-amd: I think it came to this conclusion because for the NPU1Col1 etc. objects. Maybe I can add that information instead of the 1-4

PR #89, opened 2026-03-27
Re-enable Phoenix support

andrej:
- Pick up correct kernels from aie2/aie2p subdirectories in aie_kernels where needed based off device type.
- Use the look-up-table-based operations for aie2 operators.
- Remove duplicated look-up-tables in IRON; instead use the ones that ship in the runtime lib with MLIR-AIE.
- Fix column counts and other parameters where needed for Phoenix.

Passes everything locally on a Phoenix device except for MHA (uses 8 columns, may be harder to port) and llama (which uses MHA). These are marked as "skip" on Phoenix devices using a new pytest marker.


iron/common/device_utils.py
andrej, line 1: To-do: check if there would be matching useful utilities in the IRON Python bindings in MLIR-AIE already (@hunhoffe ?)

hunhoffe: Yes there is... but I can' probably replace it in my refactor branch (which is so close to being done! So this is probably fine for now.

iron/operators/tanh/test.py
```diff
@@ -7,13 +7,18 @@
 from pathlib import Path
 
 
+from iron.common.aie_device_manager import AIEDeviceManager
 from iron.operators.tanh.op import AIETanh
+from iron.common.aie_device_manager import AIEDeviceManager
```
hunhoffe: In this file and a few others I see some double imports

hunhoffe (commented):
Nice work!

hunhoffe (approved):
Ah should have clicked approve but accidentally pressed comment the first time by mistake.

iron/common/compilation/base.py
```diff
@@ -560,6 +561,21 @@ def compile(self, artifacts):
...
         worklist = artifacts.get_worklist(KernelObjectArtifact)
         commands = []
+
+        # Select target and runtime lib directory based on device type
+        # npu1 (Phoenix/AIE2) and npu (legacy/AIE2) use aie2-none-unknown-elf
+        # npu2 (Strix Point/AIE2P) uses aie2p-none-unknown-elf
+        if self.device_type == "npu2":
+            target = "aie2p-none-unknown-elf"
+            runtime_lib_dir = "AIE2P"
+        else:
```
ypapadop-amd: I think we should throw if the device is not recognized.

iron/operators/axpy/test.py
```diff
@@ -10,10 +10,14 @@
...
 from iron.common.test_utils import run_test
+from iron.common.aie_device_manager import AIEDeviceManager
 
 
 def get_params():
-    max_aie_columns = 8
+    # Detect device and set max columns accordingly
+    # NPU1 (Phoenix) has 4 columns, NPU2 (Strix) has 8 columns
+    device_type = AIEDeviceManager().device_str()
+    max_aie_columns = 4 if device_type == "npu1" else 8
```
ypapadop-amd: I've seen this replicated a few times. Can we abstract it in dataclass that has device and max columns?

iron/operators/sigmoid/op.py
```diff
@@ -58,16 +58,37 @@ def get_mlir_artifact(self):
         )
 
     def get_kernel_artifacts(self):
-        return [
+        kernel_dir = (
+            "aie2p" if self.context.device_manager.device_str() == "npu2" else "aie2"
+        )
```
ypapadop-amd: For this too, maybe abstract it so we don't duplicate?

ypapadop-amd (changes requested):

iron/common/compilation/base.py
andrej: agreed!

iron/operators/axpy/test.py
andrej: agreed!

iron/operators/sigmoid/op.py
andrej: agreed! thanks for the good catches

ypapadop-amd (approved):
LGTM

PR #91, opened 2026-04-06
Add single-dispatch layer-by-layer multi-head attention

andrej:
"Naive" alternative implementation for multi-head attention from the currently checked-in data-flow design. This is a simple layer-by-layer implementation, but it uses the single-dispatch mechanism to fuse it all into one MLIR file and save on CPU roundtrips and XRT overheads.

Includes two variants:
1. "core": Only does the core matmuls and softmax; assumes projected and repeated inputs `Q`, `K`, `V`. This matches the functionality of the checked-in dataflow MHA.
2. "projected": Performs the Q, K, V projections, applies a RoPE positional embedding and repeats K and V matrices for grouped-query attention. Takes an embedding vector and RoPE angles as input.


iron/operators/mha_prefill_lxl/reference.py
andrej, line 1: Can we reuse the reference from the existing mha? (Note: does not include RoPE and Q, K, V projections, but some code reuse should be possible.)

aie_kernels/aie2/softmax.cc
```diff
@@ -57,13 +58,132 @@ void softmax_simple_bf16(bfloat16 *restrict input_vector, bfloat16 *restrict out
...
+// in sub-tile chunks, keeping running max and sum statistics in a small local
+// buffer (`stats`).  Layout of the stats buffer (bfloat16[16], only [0..1]
+// used):
+//   stats[0] = running max
+//   stats[1] = running sum   (of exp(x - max))
+// ---------------------------------------------------------------------------
+
+void softmax_partial_stats_impl(bfloat16 *restrict input,
+                                bfloat16 *stats,
+                                const int32_t vector_size)
```
andrej: For readability, can we make stats a struct rather than an array with hard-coded indices? Also pointer should never alias with the input so we can annotate with *restrict.

aie_kernels/aie2/softmax.cc
```diff
@@ -57,13 +58,132 @@ void softmax_simple_bf16(bfloat16 *restrict input_vector, bfloat16 *restrict out
     return;
 }
 
+// ---------------------------------------------------------------------------
```
andrej: Make sure code comment style aligns with rest of the file, not sure these `// ---` are used elsewhere

aie_kernels/aie2/softmax.cc
andrej: We'll have `struct softmax_stats *stats` as the function argument but if that's hard to  do from the `design.py` it's fine if it doesn't match from there and is just a `void *` there in the function signature

aie_kernels/aie2p/softmax.cc
```diff
@@ -159,6 +160,118 @@ void partial_softmax_alias_bf16(bfloat16 *restrict input_vector,
...
+// ---------------------------------------------------------------------------
+// Online (partial / tiled) softmax helpers
+//
+// These three kernels implement a two-pass online softmax that processes a row
+// in sub-tile chunks, keeping running max and sum statistics in a small local
+// buffer (`stats`).  Layout of the stats buffer (bfloat16[16], only [0..1]
+// used):
+//   stats[0] = running max   (scaled by log2e)
+//   stats[1] = running sum   (of exp2(x*log2e - max))
+// ---------------------------------------------------------------------------
```
andrej: Apply same comments as in aie_kernels/aie2/softmax.cc: comment style, struct instead of array for stats

aie_kernels/aie2p/softmax.cc
```diff
@@ -159,6 +160,118 @@ void partial_softmax_alias_bf16(bfloat16 *restrict input_vector,
...
+
+    auto it_in2 = aie::cbegin_restrict_vector<SM_VEC_LEN>((bfloat16 *)input);
+    for (int i = 0; i < elem_iters; i++) {
+        input_bf16 = *it_in2++;
+        scaled_accum = aie::mul(input_bf16, log2e_vec);
+        exp_in_accum = aie::sub(scaled_accum, max_val_vec);
+        aie::vector<bfloat16, SM_VEC_LEN> exp_val =
+            aie::exp2<bfloat16>(exp_in_accum.to_vector<float>());
+        exp_val_accum = add(exp_val_accum, exp_val);
+    }
```
andrej: Make this use the same on-line approach as in aie_kernels/aie2/softmax.cc

aie_kernels/aie2p/softmax.cc
andrej, line 101: There appears to be a lot of code duplicated in the new partial softmax functions below. Factor out shared functionality into a separate function used across both, forcing inlining so it does not affect performance -- or, if sharing code is not easily possible, explain what the differences between the two implementations are.

aie_kernels/generic/axpy.cc
```diff
@@ -42,4 +42,103 @@ void saxpy_scalar(bfloat16 *x, bfloat16 *y, const bfloat16 a, bfloat16 *z, const
...
+        ::aie::broadcast<bfloat16, 64>(aie::to_float<bfloat16>(a, 0));
+    for (int i = 0; i < vector_size; i += 64) {
+        ::aie::vector<bfloat16, 64> y_v = ::aie::load_v<64>(y);
+        y += 64;
+        ::aie::vector<bfloat16, 64> z_v = ::aie::add(y_v, a_v);
+        ::aie::store_v(z, z_v);
+        z += 64;
+    }
+    event1();
+}
```
andrej: Lots of needless code duplication here. Use C++ templating to create one main function that emits the multiplication/scalar add code only conditionally based on a compile-time template. Then export specializations of that function in the extern C {} section.

aie_kernels/generic/axpy.cc
```diff
@@ -42,4 +42,103 @@ void saxpy_scalar(bfloat16 *x, bfloat16 *y, const bfloat16 a, bfloat16 *z, const
...
+
+// z = (col > row) ? a : y     applied per-element of one tile, using a tile
+// position (chunk_start_col, row_in_head) supplied via the idx_buffer.
```
andrej: remove "supplied via the idx_buffer" that's a call-site implementation detail and this function might be called from other places in different ways

aie_kernels/generic/axpy.cc
```diff
@@ -42,4 +42,103 @@ void saxpy_scalar(bfloat16 *x, bfloat16 *y, const bfloat16 a, bfloat16 *z, const
...
+//
+// The tile is interpreted as a `vector_size`-wide horizontal strip of the
+// per-head (S, S) attention-score block; idx[0] is the strip's starting
+// column within that block, idx[1] is the strip's row within the block.
+// The kernel implements the causal mask in-place by writing `a` to elements
+// strictly above the diagonal and copying y -> z everywhere else.  This
+// avoids materialising an H*S*S mask buffer entirely.
```
andrej: This function is a general 'triangular fill', not specific to attention. Make the comments reflect this. Consider factoring this out into a separate operator since it has less to do with AXPY, or making it more AXPY-like by allowing a scale or constant addition be applied within the triangular masked region (again using C++ templates).

iron/common/fusion.py
```diff
@@ -1,26 +1,55 @@
...
-    """Operator that fuses multiple MLIROperators into one."""
+    """Operator that fuses multiple MLIROperators into one.
+
+    Args:
+        dispatch: Dispatch strategy for the fused operator.
+            ``"auto"`` (default) selects ``"fused"`` on NPU2 and
+            ``"separate"`` on NPU1.  ``"fused"`` uses a single-ELF
+            dispatch (requires NPU2).  ``"separate"`` compiles each
+            sub-operator to its own xclbin and invokes them sequentially.
+    """
```
andrej: Fusion's a bit of a big term for what this does. It still reconfigures the array between each operator. Let's say it fuses multiple operators into a "single dispatch." (except for NPU1 where full ELF is not yet supported where this still degenerates into multiple dispatches). The sequential feature would be used on NPU2 for debugging and benchmarking the impact of single-dispatch.

iron/common/fusion.py
```diff
@@ -52,9 +81,6 @@ def get_kernel_artifacts(self):
         ]
         for idx, op in enumerate(unique_operators):
             objs = op.get_kernel_artifacts()
-            for obj in objs:
-                obj.filename = f"op{idx}_{obj.filename}"
-                obj.prefix_symbols = f"op{idx}_"
```
andrej: Why was this removed? Disambiguating function names between the sub-operators seems desirable. An explanation / reasoning for removing it is a valid response to this if there are good reasons to remove this.

iron/common/fusion.py
```diff
@@ -82,8 +108,6 @@ def get_mlir_artifact(self):
         ]
         for idx, op in enumerate(unique_operators):
             mlir_artifact = op.get_mlir_artifact()
-            if len(op.get_kernel_artifacts()) > 0:
-                mlir_artifact.generator.kwargs["func_prefix"] = f"op{idx}_"
```
andrej: Why was this removed? Disambiguating function names between the sub-operators seems desirable. An explanation / reasoning for removing it is a valid response to this if there are good reasons to remove this.

iron/common/fusion.py
```diff
@@ -222,6 +269,58 @@ def set_up_artifacts(self):
         )
         self.add_artifacts([full_elf_artifact])
 
+    def _set_up_xclbin_artifacts(self):
+        """Chained xclbin path (NPU1/Phoenix): separate xclbin per unique operator.
+
+        Mirrors the pattern from ``chain_swiglu_artifacts`` in
+        ``iron/operators/swiglu_base.py``: each unique operator gets its own
+        xclbin + insts compiled separately, linked via ``--xclbin-input``.
```
andrej: irrelevant remove

aie_kernels/aie2/softmax.cc
```diff
@@ -57,13 +58,124 @@ void softmax_simple_bf16(bfloat16 *restrict input_vector, bfloat16 *restrict out
     return;
 }
 
+// Online (partial / tiled) softmax helpers.
+//
+// These kernels implement an online softmax that processes a row in sub-tile
+// chunks, keeping running max and sum statistics in a small per-core buffer.
+// softmax_stats names the two stats slots instead of using hard-coded array
+// indices. It occupies the first two bfloat16 elements of the stats buffer.
```
andrej:
```suggestion
```

iron/common/sequence.py
```diff
@@ -56,7 +56,9 @@ def make_callable(self, seq):
 
 
 class AutoDispatch(SequenceDispatch):
-    """Selects the platform default: full-ELF on NPU2, chained-xclbin elsewhere."""
+    """Selects the platform default: a single dispatch (full-ELF) on NPU2, and
+    a chained-xclbin dispatch elsewhere (NPU1 has no full-ELF support, so the
+    sequence degenerates into one dispatch per operator)."""
```
andrej: revert -- change is unrelated to this PR and not needed

iron/common/sequence.py
```diff
@@ -67,7 +69,11 @@ def resolve(self, device):
 
 
 class FusedDispatch(SequenceDispatch):
-    """Single-ELF dispatch (NPU2 only): all operators fused into one ELF."""
+    """Single-dispatch execution (NPU2 only): every operator in the sequence is
+    inlined into one ELF and launched with a single dispatch. The array is
+    still reconfigured between operators; "fused" refers to collapsing the
+    per-operator dispatches into one, not to merging the compute. NPU1 has no
+    full-ELF support, so it cannot use this mode (see SeparateDispatch)."""
```
andrej: revert -- change is unrelated to this PR and not needed

iron/common/sequence.py
```diff
@@ -138,6 +144,10 @@ class SeparateDispatch(SequenceDispatch):
     """Chained-xclbin dispatch: one xclbin+insts per unique operator, linked
     via ``--xclbin-input`` and invoked sequentially. Owns the compiled
     per-operator xclbin/insts maps consumed by the runtime callable.
+
+    This is the default (and only) mode on NPU1. On NPU2 it can be selected
+    explicitly to debug or to benchmark the impact of single-dispatch
+    (``fused``) execution against per-operator dispatches.
```
andrej: revert -- change is unrelated to this PR and not needed

iron/common/sequence.py
```diff
@@ -630,13 +638,19 @@ def get_buffer(self, buffer_name):
         return self._buffer_cache[buffer_name]
 
     def _sync_inputs(self):
+        # Direct host writes via get_buffer(name).data[:] don't flip the buffer's
+        # device flag, so Tensor.to("npu") would no-op (the flag reads "npu"
+        # from construction) and strand the writes on the host.  Force the
+        # host->device upload for every input buffer.
```
andrej:
```suggestion
```

iron/common/sequence.py
```diff
@@ -630,13 +638,19 @@ def get_buffer(self, buffer_name):
...
         for name in self.op.input_args:
-            self._buffers[name].to("npu")
+            self._buffers[name]._sync_to_device()
 
     def _sync_outputs(self):
+        # Force device->host for every non-input buffer so callers read fresh
+        # device results regardless of the (unchanged) device flag.
```
andrej:
```suggestion
```

PR #94, opened 2026-04-09
Add INT8 GEMM support to the GEMM operator

### LINKED ISSUE #93: INT8 GEMM support
The C++ matmul kernels (aie2/mm.cc, aie2p/mm.cc) already have vectorized INT8
  matmul templates (i8→i8, i8→i16, i8→i32, MAC shape 8x8x8) and compile flags
  (-Di8_i8_ONLY, etc.), but the Python GEMM operator only accepts bf16 input.

  This would wire up INT8 through the Python layer:

  - design.py: add "i8" to dtype_in, "i8"/"i16"/"i32" to dtype_out,
    add i8 MAC dims (8,8,8) to microkernel_mac_dim_map
  - op.py: add i8 kernel flags, min tile sizes, skip bf16 emulation for int8
  - reference.py: int8 golden reference
  - test.py: int8 test cases

  The NPU does ~50 TOPs INT8 vs ~3-5 TOPs bf16, so this would be a big
  throughput gain for quantized inference.

If I'm not missing anything it should be trivial, I'm already on it.

albiol2004:
Wire up existing INT8 matmul kernels (i8→i8, i8→i16, i8→i32) through the
  Python GEMM operator layer. The C++ kernels already had the templates and
  compile flags, this connects them to the Python API.

  Also fixes a pre-existing bug in `get_arg_spec()` where `AIERuntimeArgSpec`
  defaulted all buffers to bfloat16, causing silent data corruption for any
  non-bf16 output type.

  Closes #93

  ## Added
  - INT8 input support (`dtype_in="i8"`) with i8, i16, i32 output types
  - INT8 MAC dimensions (8,8,8) for npu1/npu2 in `microkernel_mac_dim_map`
  - INT8 kernel compilation flags (`-Di8_i32_ONLY`, etc.)
  - INT8 golden reference with int32 accumulation in `reference.py`
  - 5 INT8 test configurations (4/8 columns, all output types, row/col-major B)

  ## Changed
  - `get_arg_spec()` now passes correct `dtype` to `AIERuntimeArgSpec`
  - Test params include `dtype_in`/`dtype_out` (existing bf16 tests unchanged)
  - bf16-specific flags (prio_accuracy, bfp16 emulation) skipped for INT8

  ## Removed
  -

### DIFF
--- a/iron/operators/gemm/op.py
+++ b/iron/operators/gemm/op.py
@@ -5,7 +5,6 @@ from dataclasses import dataclass, field
 from typing import ClassVar, Dict
 
 import numpy as np
-from ml_dtypes import bfloat16
 
 from iron.common import (
     MLIROperator,
@@ -16,6 +15,7 @@ from iron.common import (
     DesignGenerator,
 )
 from iron.common.device_utils import get_kernel_dir
+from iron.common.test_utils import np_dtype_map
 import aie.utils as aie_utils
 
 
@@ -167,18 +167,9 @@ class GEMM(MLIROperator):
             ),
         ]
 
-    _np_dtype_map = {
-        "bf16": bfloat16,
-        "f32": np.float32,
-        "i8": np.int8,
-        "ui8": np.uint8,
-        "i16": np.int16,
-        "i32": np.int32,
-    }
-
     def get_arg_spec(self):
-        dtype_in_np = self._np_dtype_map[self.dtype_in]
-        dtype_out_np = self._np_dtype_map[self.dtype_out]
+        dtype_in_np = np_dtype_map[self.dtype_in]
+        dtype_out_np = np_dtype_map[self.dtype_out]
         return [
             AIERuntimeArgSpec("in", (self.M, self.K), dtype=dtype_in_np),  # input A
             AIERuntimeArgSpec(
--- a/iron/operators/gemm/op.py
+++ b/iron/operators/gemm/op.py
@@ -79,20 +79,17 @@ class GEMM(MLIROperator):
 
     @property
     def name(self) -> str:
-        """Include dtype in operator name when not the default bf16, to avoid
-        xclbin filename collisions between bf16 and int8 GEMM variants with
-        identical dimensions."""
-        base = super().name
-        if self.dtype_in != "bf16":
-            base += f"_{self.dtype_in}_{self.dtype_out}"
-        return base
+        """Include dtype in operator name to avoid xclbin filename collisions
+        between GEMM variants with identical dimensions but different dtypes."""
+        return f"{super().name}_{self.dtype_in}_{self.dtype_out}"
 
     @property
     def _kernel_flags_suffix(self):
         """Suffix encoding compile-time flags that affect the kernel binary."""
-        if self.dtype_in == "i8":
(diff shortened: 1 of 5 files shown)


iron/operators/gemm/op.py
```diff
@@ -74,9 +77,21 @@ def __post_init__(self):
...
         MLIROperator.__init__(self, context=self.context)
 
+    @property
+    def name(self) -> str:
+        """Include dtype in operator name when not the default bf16, to avoid
+        xclbin filename collisions between bf16 and int8 GEMM variants with
+        identical dimensions."""
+        base = super().name
+        if self.dtype_in != "bf16":
+            base += f"_{self.dtype_in}_{self.dtype_out}"
```
hunhoffe: I think it might make sense to always change it to include dtype in/out... thoughts @andrej ?

iron/operators/gemm/op.py
```diff
@@ -149,14 +167,29 @@ def get_kernel_artifacts(self):
...
         ]
 
+    _np_dtype_map = {
+        "bf16": bfloat16,
+        "f32": np.float32,
+        "i8": np.int8,
+        "ui8": np.uint8,
+        "i16": np.int16,
+        "i32": np.int32,
+    }
```
hunhoffe: https://github.com/amd/IRON/blob/912e6bcef9f7a225b22e17f2efe106317a375ff5/iron/common/test_utils.py#L13

albiol2004: Oops, I didn't see that one. 

Now derived from the shared one, good?

hunhoffe: Great, thanks!

albiol2004:
Found the i8 GEMM failure: `microkernel_mac_dim_map["npu1"]["i8"]` was `(8,8,8)`,
  but `aie_kernels/aie2/mm.cc` only ships a `4x8x8` wrapper. Symbol name is just
  `matmul_i8_i32` regardless of MAC dims, so link succeeds, but design.py shapes
  DMAs for `(8,8,8)` while the kernel reads them as `(4,8,8)`, producing wrong
  results. Doesn't reproduce on Strix because aie2p has the matching `8x8x8` wrapper.
  Fixed in last commit; also dropped a duplicate `"npu1"` key.

  **Long-term suggestion:** encode MAC dims in the kernel symbol
  (`matmul_i8_i32_4x8x8` instead of `matmul_i8_i32`). A Python/C++ drift would
  then become a link error instead of silent numerical garbage. Happy to do this
  in a follow-up.

iron/operators/gemm/op.py
andrej: Agreed

albiol2004: okay, done

hunhoffe:
@albiol2004 our current plan is to unify kernels in mlir-aie to avoid duplication between the two repos. This is one of the first PRs that got merged long those lines: https://github.com/Xilinx/mlir-aie/pull/3599

If it's alright with you, would you be willing to migrate your changes to kernels in mlir-aie? If not, we'll address your open PRs after we've finished with the migration to avoid churn while it's ongoing.

PR #100, opened 2026-04-09
Add scale_i32_bf16 operator (int32 to bfloat16 with per-tensor scale)

### LINKED ISSUE #99: Add scale_i32_bf16 operator
### Problem
  INT8 GEMM outputs int32 accumulators that need converting back to bf16 with a scale factor. The existing `dequant_i32_bf16` (#96) operator uses per-group packed buffer formats that don't compose directly with GEMM output. CPU-side dequantization requires expensive NPU↔CPU round trips (16MB i32 download + 8MB bf16 upload per GEMM call).

  ### Solution
  New `scale_i32_bf16` operator that takes:
  - **Input 1**: plain `(size,)` int32 buffer, directly from GEMM output
  - **Input 2**: tiny `(num_cores × 16,)` bf16 scale buffer (~256 bytes)
  - **Output**: plain `(size,)` bf16 buffer

  No packed formats. The scale ObjectFIFO is acquired once per core and reused across all tile iterations. Wins over CPU dequant at prompt lengths ≥~3500 tokens.

  ### Tests
  7 non-extensive parameter combinations, all passing. Configurations: 1-8 columns, 1-2 channels, tile sizes 256-8192.

albiol2004:
## Added
    - `aie_kernels/generic/scale_i32_bf16.cc` : C++ AIE kernel that converts int32 to bfloat16 with scalar scale multiplication. Uses the proven `aie::to_float` + `acc.to_vector<bfloat16>()` conversion pattern.
    - `iron/operators/scale_i32_bf16/` : Python operator (op.py, design.py, reference.py, test.py).

  ## Changed
    Nothing. This is a new operator with no modifications to existing code.

  ## Removed
    Nothing.

  ## Motivation
    The existing `dequant_i32_bf16` operator uses per-group packed buffer formats (`[i32_data | bf16_scales]` as uint8) that don't compose directly with INT8 GEMM output. CPU-side dequantization requires expensive NPU↔CPU round trips (16MB i32 download + 8MB bf16 upload per GEMM call).

    This operator takes plain typed buffers, int32 input directly from GEMM, a tiny bf16 scale buffer (~256 bytes), and bf16 output, enabling fully on-NPU i32→bf16 dequantization with zero CPU round trips.

    Key design: the scale ObjectFIFO is acquired once per core and held across all tile iterations, so the scale buffer stays minimal (num_cores × 16 bf16 values) regardless of tensor size. Wins over CPU dequant at prompt lengths ≥ ~3500 tokens.

  ## Test results
    - 28/28 parametrized tests pass (input lengths 1K–8K, 1–8 columns, 1–2 channels, tile sizes 128–8192)
    - rel_tol=0.01, abs_tol=1e-6 (matches dequant_i32_bf16 tolerances)
    - No regressions in existing operators

  ## Checklist
    - [x] Tests pass locally (28/28 including extensive)
    - [x] Code formatted (black + clang-format)
    - [x] License headers on all files
    - [x] No changes to existing code

  Closes #99

### DIFF
--- a/aie_kernels/generic/scale_i32_bf16.cc
+++ b/aie_kernels/generic/scale_i32_bf16.cc
@@ -1,0 +1,31 @@
+// SPDX-FileCopyrightText: Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <aie_api/aie.hpp>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+// Convert int32 to bfloat16 with scalar scale multiplication.
+// scale_tile contains 16 bfloat16 values; only the first is used (broadcast).
+// out[i] = bf16(float(in[i]) * float(scale_tile[0]))
+extern "C" {
+
+void scale_i32_to_bf16(int32_t *restrict in, bfloat16 *restrict scale_tile, bfloat16 *restrict out, int32_t size)
+{
+    float scale_f32 = (float)scale_tile[0];
+    aie::vector<float, 16> scale_vec = aie::broadcast<float, 16>(scale_f32);
+
+    event0();
+    for (int i = 0; i < size; i += 16) {
+        aie::vector<int32_t, 16> i32_vec = aie::load_v<16>(in + i);
+        aie::vector<float, 16> f32_vec = aie::to_float<float>(i32_vec, 0);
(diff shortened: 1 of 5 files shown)


albiol2004:
Overhead of using it as a standalone operator is so big that kills gains in the Llama pipeline, but served my research purposes. I don't know if it may be useful to keep it...

PR #79, opened 2026-03-07
Add decode dataflow fusion operators

jgmelber:
## Summary
- Implements 4 new fused operators for the decode dataflow pipeline: `swiglu_fused_decode` (Phase 3), `fused_qkv_proj` (Phase 1), `flowkv_decode` (Phase 2), and `fused_dequant_gemv` (Phase 5)
- Eliminates DDR round-trips for intermediate activations by keeping data on-chip via inter-tile ObjectFIFOs
- Fixes static buffer overflow in `dual_gemv_silu_mul` kernels (1024 → 2048 elements) and adds production-scale test parameters

## Test plan
- [x] `dual_gemv_silu_mul` — 5/5 passed (buffer overflow fix verified on hardware)
- [x] `fused_qkv_proj` — 5/5 passed (reuses `mv.o`, Llama dims M=3072 K=2048)
- [x] `flowkv_decode` — 5/5 passed (online softmax, 2-tile pipeline per KV head group)
- [x] `fused_dequant_gemv` — 5/5 passed (INT4→bf16 fused dequant+GEMV)
- [x] `swiglu_fused_decode` — 5/5 passed (2-stage pipeline, 1.32x speedup at Llama dims)
- [x] Run `black --check .` and `reuse lint`
- [ ] Run extensive tests at Llama 3.2 1B production dimensions (2048×8192)

Hardware: AMD Ryzen AI 9 HX 370 (RyzenAI-npu4), XRT 2.21.75

### Benchmark: `swiglu_fused_decode` vs `swiglu_decode` (Llama 3.2 1B dims)

| Metric | Baseline (2 runlists) | Fused (1 runlist) | Improvement |
|---|---|---|---|
| Median latency | 5410 µs | 4103 µs | **1.32x** |
| Effective bandwidth | 18.61 GB/s | 24.54 GB/s | **1.32x** |
| DDR intermediate | 32 KB round-trip | 0 KB (on-chip) | Eliminated |
| 16-layer MLP tok/s | 11.6 | 15.2 | **+31%** |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #101, opened 2026-04-09
Add GEMV INT 4 

albiol2004:
## Added
  - `aie_kernels/generic/fused_dequant_gemv.cc` : C++ AIE kernel that fuses INT4 weight dequantization with matrix-vector multiplication in a single pass. Uses the proven `aie::unpack` chain from `expand.cc` (uint4 → uint8 → uint16 → bf16 → scale → MAC). Includes a double-pump optimization that processes 2 groups per iteration for instruction interleaving, and compile-time `DIM_K`/`GROUP_SIZE` for loop specialization.
  - `iron/operators/gemv_int4/` : Python operator (op.py, design.py, reference.py, test.py) following the existing `gemv/` and `dequant/` patterns. Packed weight buffer layout: `[uint4 weights | bf16 per-group scales]` per tile.

  ## Changed
  Nothing. This is a new operator with no modifications to existing code.

  ## Removed
  Nothing.

  ## Motivation
  Decode inference is bandwidth-bound, GEMV loads each weight once per token. INT4 weights are half the size of INT8 and a quarter of bf16, making INT4 GEMV the highest-impact single operator for decode throughput.

  This is a **foundation operator**: a standalone, individually tested building block for INT4 weight inference. It is not intended as an optimized fused decode kernel, the goal is to provide a correct,
  well-tested INT4 GEMV that can be composed into fused pipelines.

  Credit to @jgmelber for the original fused dequant-GEMV concept in PR #79 and the INT4 GEMV benchmarks in PR #71 (21.4 GB/s on 8192×2048). That work demonstrated the viability of fused INT4 inference on AIE and informed the design here. If you have insights on further kernel optimizations I'd love to hear them, my focus was on getting the foundation block right rather than a fully optimized fused kernel.

  ## Test results
  - 7/7 parametrized tests pass (2048×2048, 8192×2048, 2048×8192 at 4/8 columns, tsi=1/4)
  - Integration tested with real Llama 3.2 1B weights (cosine similarity >0.999 vs CPU reference)
  - Performance (8 columns): 2048×8192 at 561 us / 16.9 GB/s, 8192×2048 at 665 us / 14.2 GB/s

  ## Checklist
  - [x] Tests pass locally
  - [x] Code formatted (black + clang-format)
  - [x] License headers on all files
  - [x] No changes to existing code

### DIFF
--- a/iron/operators/gemv_int4/design.py
+++ b/iron/operators/gemv_int4/design.py
@@ -1,0 +1,160 @@
+# SPDX-FileCopyrightText: Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Fused INT4 dequantization matrix-vector design.
+
+Performs a fused dequantize-GEMV where the weight matrix is stored in packed
+INT4 format (two 4-bit values per uint8 byte) with per-group bfloat16 scale
+factors.  The activation vector and output are bfloat16.
+
+Each AIE column processes a contiguous block of output rows.  Within a column,
+the worker iterates over tiles of packed weight rows, acquires the full
+activation vector once per outer iteration, and calls the fused dequant-matvec
+kernel which unpacks, dequantizes, and accumulates in a single pass.
+
+Buffer layout for A (packed weights, uint8):
+    For each tile of m_input rows: [m_input * K / 2 bytes of packed weights]
+                                   [m_input * (K / group_size) * 2 bytes of scales]
+"""
+
+import numpy as np
+from ml_dtypes import bfloat16
+
+import aie.dialects.index as index
+from aie.dialects.aie import T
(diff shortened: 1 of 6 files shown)


iron/operators/gemv_int4/design.py
```diff
@@ -0,0 +1,161 @@
+# SPDX-FileCopyrightText: Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
```
hunhoffe, line 1: How much of this is similar to other GEMM ops? Can we find a way to reuse code?

albiol2004: Like 70% similar to gemv. But different layout and kernel. I could import from gemv to reuse _build_gemv_program()


or have it as a common helper (only 2 ops using it I think) which I would leave for refactor in another PR


Basically it’s the fused kernel that is important so depends on what you prefer for readability/maintenance

PR #105, opened 2026-04-14
Fix swiglu_decode intermediate comparison to use chained reference

albiol2004:
Aligns `swiglu_decode/test.py` with `swiglu_prefill/test.py` by verifying the `intermediate` buffer against a chained reference built from the observed AIE `left_swished` and `right` buffers, instead of against the CPU-computed `golden_ref["intermediate"]`.

  The golden-reference path amplifies legitimate, sub-tolerance bf16 drift from upstream stages (e.g. SiLU of very-negative inputs where the AIE LUT rounds to `0.0` while fp32 CPU silu preserves a tiny negative value) through the multiplication against a large-magnitude `right` operand, producing spurious "got 0.0, expected -1.27"-style failures. The AIE kernels themselves are numerically correct, the observed `intermediate` matches `observed_left_swished * observed_right` exactly, and the final `output` already passes at a tighter tolerance than the intermediate stage.

  This issue surfaces at rectangular FFN shapes (e.g. `embedding_dim=1024, hidden_dim=3584`) where the statistics of the SiLU input distribution make near-zero LUT outputs more common than at the previously-tested square `2048²` shape. Adds `(1024, 3584)` to the parametrization so regressions in rectangular decode are caught.

  ## Added
  - `(1024, 3584)` parametrization in `iron/operators/swiglu_decode/test.py`, reflecting Qwen3.5-0.8B FFN dims so rectangular decode is covered alongside the existing square smoke test.

  ## Changed
  - `iron/operators/swiglu_decode/test.py`: verify `intermediate` against a chained reference (`observed_left_swished * observed_right`) rather than `golden_ref["intermediate"]`, matching the approach already in `swiglu_prefill/test.py`. Tightens the tolerance to `rel_tol=0.04, abs_tol=0.4` accordingly (same values used for the output check and for prefill).

  ## Removed
  - None.

  ## Testing
  Verified on NPU2 (Strix, `aie2p`):

  - `pytest iron/operators/swiglu_decode/test.py -v --iterations 1` : both `2048×2048` and `1024×3584` pass.
  - `pytest iron/operators/ -m "not extensive" --iterations 1` : no regressions.

  ## PR Merge Checklist

### DIFF
--- a/iron/operators/swiglu_decode/test.py
+++ b/iron/operators/swiglu_decode/test.py
@@ -13,7 +13,14 @@ from iron.common.test_utils import verify_buffer
 
 
 def get_params():
-    params_list = [(2048, 2048)]
+    # (embedding_dim, hidden_dim)
+    # Square shape is the historical smoke-test config; the rectangular
+    # shape reflects real decoder-model FFN dims (e.g. Qwen3.5-0.8B
+    # embedding=1024, hidden=3584) that downstream runtimes actually hit.
+    params_list = [
+        (2048, 2048),
+        (1024, 3584),
+    ]
 
     params = []
     for p in params_list:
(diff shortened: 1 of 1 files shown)


andrej:
Thanks, this is a useful improvement. I just kicked off the CI, if this passes I'll merge it

andrej (approved):

PR #107, opened 2026-04-14
Use --find-links for MLIR-AIE and LLVM-AIE wheel sources

x4v13r64:
The GitHub Release URLs for `mlir_aie` and `llvm-aie` are currently added as `--extra-index-url` entries alongside PyPI. This means pip queries all indexes for all packages, which can cause resolution issues when the same package name exists on multiple indexes at different versions. Switching to `--find-links` adds the wheel files as candidates without creating a competing package index.


andrej (approved):
This is a good improvement -- sorry I'm only now getting to it. Will merge this ASAP.

PR #111, opened 2026-05-04
Add `xchesscc` as an alternative kernel compiler option

andrej:
Used this to debug the `alignas` bug, might be useful for other things where there's performance or functionality differences between peano and chess


hunhoffe (approved):
Sorry this slipped through the cracks, but this looks good to me!

Issue #113, opened 2026-05-11
Re-enable tile-size-8192 GeLU test


andrej:
I will disable the 8192-tile-size GeLU test for Phoenix that I recently added because it currently causes a red X for the extensive test suite. Root cause is the issue fixed with [MLIR-AIE PR #3045](https://github.com/Xilinx/mlir-aie/pull/3045). Once we merge that in the test can be re-enabled.

PR #114

andrej:
Closed by #141

PR #114, opened 2026-05-11
Temporarily disable ReLU test, revert llvm-aie version change, reformat CI output

andrej:
The CI was down for a while. Now that it ran again it revealed the following issues which this PR fixes:

- GeLU extensive test for tile size 8192 was failing. Fix will be in [MLIR-AIE PR 3045](https://github.com/Xilinx/mlir-aie/pull/3045) -- until then, disable the test (tracking with issue #113).
- llvm-aie version cannot be found. Revert to previous peano.
- PR #88 made it so that all tests are only named by their parametrization (e.g. M_8192_K_8192...) so it becomes unclear what the actual tested operator is (e.g. GeLU). Include operator names.
- Reformat CI summaries a bit more pretty-ly.

### DIFF
--- a/iron/applications/llama_3.2_1b/llama_npu.py
+++ b/iron/applications/llama_3.2_1b/llama_npu.py
@@ -1245,7 +1245,7 @@ def llama_forward_pass_decode(config, state):
     assert seq_len == 1
     assert state.num_preceding_tokens < max_seq_len
 
-    # patch_fused_decode_operator(aie_ops.decode, config, state.num_preceding_tokens)
+    patch_fused_decode_operator(aie_ops.decode, config, state.num_preceding_tokens)
 
     # Prefill RoPE angle look-up tables
     angles_slice = config.angles[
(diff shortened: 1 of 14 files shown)


hunhoffe (approved):

iron/applications/llama_3.2_1b/llama_npu.py
```diff
@@ -1245,7 +1245,7 @@ def llama_forward_pass_decode(config, state):
     assert seq_len == 1
     assert state.num_preceding_tokens < max_seq_len
 
-    patch_fused_decode_operator(aie_ops.decode, config, state.num_preceding_tokens)
+    # patch_fused_decode_operator(aie_ops.decode, config, state.num_preceding_tokens)
```
jgmelber: Should this be left in or deleted?

jgmelber (approved):

iron/applications/llama_3.2_1b/llama_npu.py
andrej: Great catch, needs to be reverted actually because the patching is still necessary.

PR #115, opened 2026-05-11
Update MLIR-AIE Wheel Dependency to use new Placer Logic

andrej:
Some downstream work I have in the queue depends on a newer compiler version and but that comes with the new placer logic. It's cleaner to do the upgrade in a separate PR so proposing to do it now.

### DIFF
--- a/iron/common/compilation/base.py
+++ b/iron/common/compilation/base.py
@@ -701,6 +701,7 @@ class KernelCompilationRule(CompilationRule):
                         "-O2",
                         "-std=c++20",
                         f"--target={target}",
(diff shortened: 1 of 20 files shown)


hunhoffe (approved):
Whoops, another PR I missed! Thanks for doing this. Definitely move forward if this is blocking you. I had planned to do the minor release + IRON upgrade once the following landed:
* https://github.com/Xilinx/mlir-aie/pull/3104
* https://github.com/Xilinx/mlir-aie/pull/3099
* https://github.com/Xilinx/mlir-aie/pull/3078

However, I'm also more than fine with an intermediate update landing first, I can do the next update later :)

PR #117, opened 2026-05-12
Enable separate-xclbin dispatch for FusedMLIROperator for debugging and Phoenix support

andrej:
While single-dispatch operators improve performance, it makes it hard to debug when something goes wrong. This adds several modes to the `FusedMLIROperator` to be able to dispatch layer-by-layer as separate xclbins (vs. the previous single full ELF dispatch), inspect outputs after each layer and compare against a reference for troubleshooting. This also enables Phoenix support, which does not have full ELF, by falling back to the same layer-by-layer xclbin dispatch.

Lastly, this also renames the poorly named `FusedMLIROperator` to `OperatorSequence`.


iron/common/sequence.py
```diff
@@ -0,0 +1,815 @@
...
+            args_specs = op.get_arg_spec()
+            if len(args_specs) != len(bufs):
+                raise ValueError(
+                    f"Number of buffers ({len(bufs)}) must match operator argument "
+                    f"specification ({len(args_specs)}) for operator {op!r}"
+                )
+            for i, buf_name in enumerate(bufs):
+                args_spec = args_specs[i]
+
+                # Parse slice notation: "buffer_name[start:end]"
```
hunhoffe, line 320: Do we really need our own notation? Can we use a structured format like json or something?

iron/common/compilation/sequence.py
```diff
@@ -32,7 +32,7 @@
 # ##########################################################################
 
 
-class FusedMLIRSource(CompilationArtifact):
+class SequenceMLIRSource(CompilationArtifact):
```
hunhoffe, line 35: Does not match the PR description? I like either new name though.

jgmelber (approved):

iron/common/utils.py
```diff
@@ -94,4 +130,27 @@ def from_parent(cls, parent, shape, offset_elements, length_elements, dtype):
             size_bytes=length_elements * itemsize,
             shape=shape,
             dtype=dtype,
+            parent=parent,
         )
+
+
+class CPUBuffer:
```
hunhoffe: Can you use `CPUOnlyTensor` for this? https://github.com/Xilinx/mlir-aie/blob/75219495817e47cadb9e010c13048330b3af7357/python/utils/hostruntime/tensor_class.py#L668

iron/operators/elementwise_mul/reference.py
```diff
@@ -5,11 +5,15 @@
 from iron.common.test_utils import torch_dtype_map
 
 
+def reference(a, b):
```
hunhoffe, line 8: Nice! This is getting closer to the format I put the programming examples in in mlir-aie -- I'd like to continue unifying, maybe backport this reference structure to mlir-aie at some point: https://github.com/Xilinx/mlir-aie/blob/75219495817e47cadb9e010c13048330b3af7357/programming_examples/basic/row_wise_bias_add/row_wise_bias_add.py#L114

iron/operators/gemv/op.py
```diff
@@ -99,3 +99,9 @@ def get_arg_spec(self):
             AIERuntimeArgSpec("in", batch_dim + (self.K,)),  # vector
             AIERuntimeArgSpec("out", batch_dim + (self.M,)),  # output
         ]
+
+    def reference(self, A, B):
+        """CPU reference: (optionally batched) matrix-vector product."""
+        from iron.operators.gemv.reference import reference
```
hunhoffe, line 105: Is it really worth keeping the reference in a separate file?

iron/tests/infrastructure/sequence.py
```diff
@@ -0,0 +1,259 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Infrastructure tests for :class:`OperatorSequence`.
```
hunhoffe, line 5: Nice! Re: unification, someday would be cool to get this ported to the runlist/sequencing capabilities in mlir-aie too!

hunhoffe (approved):
Good work! A few comments, but nothing blocking.

iron/operators/gemv/op.py
andrej: Since all operators already had their reference in a standalone file, if they had any, I just went with that here, but I agree that it probably wouldn't need to be in its own file. I guess it doesn't hurt either though, if any of these operators did any sort of complicated calculations, that way we have a clear distinction, as in `op.py` is only operator metadata, `design.py` is the NPU implementation and `reference.py` is the CPU implementation.

hunhoffe: I agree, no harm. I just like simplicity when possible :)

iron/common/compilation/sequence.py
andrej: They're actually not the same thing. There's both the `OperatorSequence`, defined under `iron/common/sequence.py` -- that one is the operator metadata, describing the sequence of operators you want to call -- and this `SequenceMLIRSource`, which is just an auto-assembled MLIR source that 'concatenates' multiple input MLIR operators into one long sequence using `aiex.configure` and `aiex.run`. 

The connection between the two: If you use the `'fused'` dispatch mode, the operators in the `OperatorSequence` will be concatenated together into one `SequenceMLIRSource`  -- a single compilation artifact/source for the entire sequence. However, if you chose one of the other dispatch modes, `SequenceMLIRSource` is not used, and instead each individual operator is compiled and executed.

iron/common/sequence.py
andrej: Not changed by this PR (though it shows up in the diff because I renamed files and the diff is so large that it looks like a new file) but open to the idea.

For context, [here](https://github.com/amd/IRON/blob/95476293949becf9a4579f8e13c98c5f10b87333/iron/applications/llama_3.2_1b/llama_npu.py#L516-L521) is how this notation looks like at one of the usage sites. I actually quite hate it, with it being in a string like that. I guess when I first experimented with it, the runlists looked much more toy-like, so having the notation was nice, but now that the slices are computed in Python, then concatenating them back into the string is ugly. Just a tuple (buffer name, start, end) would probably be best.

iron/tests/infrastructure/sequence.py
andrej: Agreed, but it has to live in IRON for now because it relies on our operator metadata that the `op.py`s describe, especially for the argument specifications.

The rest of the infrastructure that can live in MLIR already does (`aiex.configure`, `aiex.run`).

iron/common/utils.py
andrej: Yes! Extremely good catch

iron/common/compilation/sequence.py
hunhoffe: Oh I see! Thank you for explaining.

iron/common/sequence.py
hunhoffe: Yeah I think maybe a tuple would make sense (if you get really fancy you could used a NamedTuple but IIRC their performance was relatively poor compared to the tuple at scale). But, not too important at this point.

Issue #54, opened 2025-12-10
Intermittent CI failures for llama on later iterations


andrej:
#44 temporarily disabled running llama for multiple iterations because the subsequent iterations would intermittently fail.

The issue was not reproducible on `xcoradaie219` and only happens intermittently on the CI machine.

To-Do: investigate

<img width="1047" height="194" alt="Image" src="https://github.com/user-attachments/assets/e8b4a626-4630-4a30-becf-bdf8663bcb89" />
<img width="1236" height="148" alt="Image" src="https://github.com/user-attachments/assets/75a30813-ee22-471d-afda-271bdff8ebcf" />

theonlychant:
you can check the PR I made when you're free I looked into this [PR](https://github.com/amd/IRON/issues/54)

PR #118, opened 2026-05-13
fix: restore multi-iteration llama runs with intermittent failure guard

theonlychant:
Addresses #54

Restores multi-iteration llama test runs that were temporarily disabled
in #44 due to intermittent CI failures. The failures were not
reproducible on xcoradaie219 and only occurred on the CI machine.
Root cause has been identified and resolved in iron/common/test_utils.py.

## Added
-

## Changed
- Restored multi-iteration llama execution in iron/common/test_utils.py
  with fix for intermittent CI failure condition

## Removed
- Temporary disable/workaround introduced in #44
image
image

### DIFF
--- a/iron/common/test_utils.py
+++ b/iron/common/test_utils.py
@@ -84,6 +84,38 @@ def verify_buffer(
     expected_np = _to_numpy(reference).reshape((-1,))
     output = _to_numpy(output).reshape((-1,))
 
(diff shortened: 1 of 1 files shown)


andrej:
This PR apparently adds some debug logging/storing outputs to disk on failure. Looking at the changeset I cannot see anything matching the description of the PR that fixes the issue. Maybe you accidentally forgot to commit some changes? If so feel free to reopen once updated.

PR #121, opened 2026-06-16
Register design modules in sys.modules to fix path wrangling (#42)

### LINKED ISSUE #42: Clean up test discovery and operator paths
After #37, I will need to

- [x] Replace the "retrofitted" self-baked test discovery script with something established and simpler (pytest?)
- [ ] Remove the path wrangling from `compilation.py` -- the `design.py`s should be available in the Python path

masteryi-0018:
Close #42

## Summary

Remove the path wrangling in `compilation.py` by registering design modules in `sys.modules` under their fully-qualified package names.

### Before
- `DesignGenerator.__call__` loaded design modules with `spec_from_file_location(filename, path)` using only the file basename as the module name -- the module was not registered, so code that needed the design later had to duplicate the same `spec_from_file_location` dance.
- `get_child_mlir_module` in `fusion.py` duplicated the same 3-line `spec_from_file_location` + `module_from_spec` + `exec_module` pattern.

### After
- `DesignGenerator.__call__` derives a proper dotted module name from the file path (e.g. `iron.operators.axpy.design`), registers the loaded module in `sys.modules`, and loads it.
- `get_child_mlir_module` now simply calls `importlib.import_module(gen.module_name)` to retrieve the already-registered module.

### Changes

| File | Change |
|---|---|
| `iron/common/compilation/base.py` | + `module_name` property, + `sys.modules[module_name] = module` |
| `iron/common/compilation/fusion.py` | 3-line spec_from_file_location → 1-line importlib.import_module |

Zero changes to operator callers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/iron/common/compilation/base.py
+++ b/iron/common/compilation/base.py
@@ -56,18 +56,39 @@ from iron.common.device_utils import get_kernel_dir
 
 @dataclass
 class DesignGenerator:
-    """Lazy callable that imports source_path and calls fn_name(*args, **kwargs), returning MLIR as a string."""
+    """Lazy callable that imports *source_path* and calls *fn_name*, returning MLIR.
+
+    On first call the loaded module is registered in ``sys.modules`` under
+    its fully-qualified package name (e.g. ``iron.operators.axpy.design``) so
+    that subsequent code can import it through the standard Python import
+    mechanism.
+    """
 
     source_path: Path
     fn_name: str
     args: tuple = ()
     kwargs: dict[str, Any] = field(default_factory=dict)
 
+    @property
+    def module_name(self) -> str:
+        """Python module path derived from *source_path*.
+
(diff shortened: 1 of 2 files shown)


andrej (commented):
Thank you for your contribution, and my sincere apologies it took me so long to give this the attention it deserved.

After looking over the code, I am not convinced that this is a meaningful improvement for clarity. I'm happy to hear arguments as to how this is better, but from my perspective, adding the import path to `sys.modules` in one place, with the relative path-ancestor logic, and then relying on `sys.modules` being set in another place, where it is imported, is less clear than the explicit importlib logic.

I will close the referenced issue #42.

Please don't let this discourage you from further contributions. We really value your engagement and I will try my best to be more responsive in the future.

PR #122, opened 2026-06-18
Add stream-dse fused SwiGLU-prefill operator

asyms:
Adds `SwiGLUPrefillStream`, a SwiGLU-prefill operator whose MLIR design (gate/up GEMMs + SiLU + elementwise-mul + down GEMM) is generated at build time by [stream-dse](https://github.com/KULeuven-MICAS/stream) and deployed as one full ELF, instead of chaining separately compiled sub-operators.

Both inputs stream-dse needs come from IRON. The workload is `torch.onnx.export` of the reference `nn.Module` the test checks against, so the deployed design cannot drift from what is verified. The mapping is emitted from a placement written in array columns, with node names taken from that same export, so workload and mapping cannot disagree. Neither file is committed; both are written into the build directory.

`k` selects how many fused groups the block is split into, and so how many designs the ELF holds:

| k | designs | arrangement |
|---|---|---|
| 1 | 1 | whole block resident on the array |
| 2 | 2 | gate/up/SiLU/mul, then the down projection |
| 5 | 5 | layer by layer, each layer taking the whole array in turn |

k=5 is the arrangement `swiglu_prefill` uses with IRON's hand-written operators, so the two flows are directly comparable.

## Added

- `iron/operators/swiglu_prefill_stream/`: the operator, `k` in {1, 2, 5}. `reference.py` is the module the design is generated from and names everything downstream; `stream_design.py` builds the workload and mapping and calls stream-dse.
- `iron/common/stream/`: the reusable integration. `workload.py` exports a module to ONNX, `mapping.py` emits the mapping from a placement, `hardware.py` resolves array columns to stream core ids via mlir-aie's `Device`, `ops.py` binds each layer to IRON's existing `aie_kernels` sources.
- `iron/common/layout.py`: `TiledStridedLayout`, the operand layouts the generated DMAs and the compiled kernels must agree on.
- `iron/tests/stream/`: layouts match stream-dse's, names come from the golden reference, placements resolve to the expected cores. No NPU required.
- `requirements_stream.txt`: optional dependency `stream-dse>=1.13.9`; the operator's test skips when it is absent.

## Changed

- `iron/common/sequence.py`: `OperatorSequence` forwards `extra_flags` into the full-ELF build, so a placed and routed whole-array design can pass `--dynamic-objFifos`.

## Removed

- None.

## Running it

```bash
source /opt/xilinx/xrt/setup.sh            # XRT on PATH (pyxrt + xclbinutil)
pip install -r requirements.txt            # IRON + mlir_aie/llvm-aie + torch
pip install -r requirements_stream.txt     # stream-dse (PyPI)
stream-setup-aie                           # installs stream-dse's AIE codegen dialects
pytest iron/operators/swiglu_prefill_stream/
```

Runs k=1, k=2 and k=5. Each exports the workload, emits the mapping, solves the allocation, generates and compiles the design, verifies against the golden reference and reports latency.

Outputs land in `outputs/<hardware>-swiglu[_k<k>]_<m>_<e>_<h>-<rows>_row_<cols>_col/`: `workload.onnx`, `mapping.yaml`, and the generated MLIR under `codegen/` (k=1) or `group_i/codegen/` (k>1). The ELF and kernel objects go to `build/`.

Measured on Strix (npu2), seq 256 / embedding 512 / hidden 2048, mean of five dispatches: k=1 2465 us, k=2 3216 us, k=5 3262 us. Fusing the whole block is worth 30% against either split; splitting past the first cut costs about 1%.

## Notes

- Design generation uses the open-source OR-Tools GSCIP solver, so no Gurobi license is needed. `stream-setup-aie` is required and leaves IRON's pinned `mlir_aie`/`llvm-aie` alone.
- Importing the operator does not require stream-dse; only building it does.
- Every dispatch takes a fresh callable. The full-ELF path does not reset array state between dispatches, so reusing one reads stale state and a design of several fused groups deadlocks on the second dispatch. Characterised and raised separately.
- `iron/tests/` is not run by any CI workflow, since every workflow passes an explicit test path, so `iron/tests/stream/` needs running by hand.

## Licensing note

Files this PR creates carry a `KU Leuven (MICAS)` copyright header (Apache-2.0), as they were authored by MICAS: `iron/common/layout.py`, `iron/common/stream/*`, `iron/operators/swiglu_prefill_stream/*`, `iron/tests/stream/*` and `requirements_stream.txt`. Files this PR only modifies keep their existing AMD headers. Happy to discuss.

## PR Merge Checklist


andrej:
Hi Arne, sorry for the CI failures, if you rebase on #125 maybe once it's merged hopefully these should pass

iron/operators/swiglu_prefill_stream/test.py
```diff
@@ -0,0 +1,103 @@
...
+
+import time
+import pytest
+
+# The fused design is generated by stream-dse at compile() time. stream-dse is an
+# optional dependency (see requirements_stream.txt) that the default CI image does
+# not install, so skip this whole module when it is unavailable.
+pytest.importorskip(
+    "stream", reason="stream-dse not installed (see requirements_stream.txt)"
+)
```
andrej, line 15: Could you please add `requirements_stream.txt` [here](https://github.com/amd/IRON/blob/4d4b803c99fe896d0f08aba5dcc86a8f319c18a5/.github/actions/prereqs/action.yaml#L23) so that this test runs in CI?

andrej (changes requested):

asyms:
@andrej Could you re-run the workflows? I have rebased on top of `devel` and fixed the timeout issue (I think).

.github/actions/prereqs/action.yaml
```diff
@@ -21,4 +21,9 @@ runs:
         source ${{ inputs.env_name }}/bin/activate
         pip install --upgrade pip
         pip install -r requirements.txt
+        pip install -r requirements_stream.txt
+        # stream-dse's AIE codegen deps (snax-mlir/snaxc, xdsl-aie, aie-python-extras)
+        # are not PyPI dependencies; this console script installs them so the
+        # stream-backed operators can generate their MLIR at build time.
+        stream-setup-aie
```
andrej: I think this might potentially be the source of CI failures. If `stream-setup-aie` installs aie-python-extras, it might overwrite the version of those extras installed by the mlir-aie wheel (which is now called EUDSL I believe).

iron/common/compilation/sequence.py
```diff
@@ -89,6 +92,26 @@ def get_child_mlir_module(mlir_artifact: PythonGeneratedMLIRArtifact) -> Any:
...
+    makes the count even. Consecutive entries running the same operator share a
+    configure point.
+    """
+    points = 0
+    previous = None
+    for op_name, *_ in runlist:
+        if op_name != previous:
+            points += 1
+            previous = op_name
+    return points % 2 == 1
```
andrej, line 112: Good for now, opened issue #143 to get a long-term fix into MLIR-AIE

andrej (approved):

PR #123, opened 2026-06-18
gemm: honor func_prefix for kernel object and symbols (enables fusion)

atassis:
GEMV, softmax, and binary_elementwise already apply `func_prefix` to both the kernel object filename and the kernel symbols, so `FusedMLIROperator` can place several of them in one design. GEMM only applied `func_prefix` to the fallback generated-object name, leaving the supplied `kernel_object` name and the `zero_`/`matmul_`/`convert_copy_` symbols unprefixed — so fusing any GEMM via `FusedMLIROperator` failed with object-copy and undefined-symbol errors. (The in-tree `llama_3.2_1b` example builds a `FusedMLIROperator`.)

**Changed**
- `my_matmul` now applies `func_prefix` to the gemm and `convert_copy` object filenames and to the `zero_`/`matmul_`/`convert_copy_` kernel symbols.

**Added / Removed**
- None.

No-op when `func_prefix=""` (the non-fused default), so existing callers are byte-identical. Exercised in our own batched decode, which fuses GEMM via `FusedMLIROperator`.

### DIFF
--- a/iron/operators/gemm/design.py
+++ b/iron/operators/gemm/design.py
@@ -280,7 +280,11 @@ def my_matmul(
 
     # AIE Core Function declarations
     scalar_suffix = "_scalar" if use_scalar else ""
-    gemm_object = kernel_object or f"{func_prefix}gemm_{m}x{k}x{n}.o"
+    gemm_object = (
+        f"{func_prefix}{kernel_object}"
+        if kernel_object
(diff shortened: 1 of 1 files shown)


jgmelber (approved):

andrej:
Note the CI failures we're seeing should disappear after a rebase. :)

PR #127, opened 2026-06-26
gemv: coalesce batched DMA into a single iterated BD per column

atassis:
Replaces the GEMV operator's per-batch host DMA descriptor unroll with a single iterated descriptor per column.

GEMM already folds many transfers into one multi-dimensional iterated descriptor (via `TensorTiler2D`); GEMV did not, so a batched GEMV emitted `num_batches` separate fill/drain descriptors (plus a per-batch task-group `wait`) per AIE column, linear in the batch count. The test suite also had no `num_batches > 1` GEMV coverage, so the batched path shipped untested.

## Added
- A `num_batches > 1` golden reference (`generate_golden_reference_batched`) and a parametrized device test (`test_gemv_batched`) covering: the coalesced path with large `num_batches` (the size-uncapped descriptor dimension) and a multi-dimension run split; a run that requires an alignment-aware (even) inner split; the per-batch fallback (a batch stride over the limit); and an attention-style shape with `tile_size_input > 1` and `num_batches` = a head count.

## Changed
- The GEMV runtime now coalesces the batched A-fill / C-drain into one iterated descriptor per column by default. `num_batches` is placed in the size-uncapped descriptor dimension and the contiguous per-batch run is split across the two wrap dimensions (each <= 1023), per the AIE shim BD limits (`AIEXDialect.cpp` `verifyStridesWraps`). The run split keeps the innermost size and the strides 4-byte-aligned (the shim's address granularity, which is enforced even for linear transfers).
- The per-batch drain `wait` is replaced by ObjectFifo backpressure. Depth 2 is sufficient (rather than depth `num_batches`): the A-fill (MM2S) and C-drain (S2MM) run on separate shim channels and the pipeline is paced by the fifo locks, so the producer cannot overrun the consumer; this is a streaming pipeline, not a store-all buffer. The depth invariant is asserted.
- The coalesced descriptor accesses the exact same DRAM elements in the same order as the previous unroll. `num_batches == 1`, and any configuration that cannot be coalesced (a run with no aligned split under the wrap cap, or a batch stride that is too large or not aligned), fall back to the existing per-batch path and are byte-identical. The coalesced path therefore introduces no new failure modes relative to the unroll.

## Removed
- Nothing (behavior for existing callers is unchanged).

## Motivation and measurements
This is a descriptor-count / build-time and correctness change, not a runtime speedup: the transfer is access-equivalent (same bytes, same order), so it does not change data-movement time.

The descriptor count for a batched GEMV drops from O(num_batches) to O(1) per column, which shrinks the descriptor-bound lowering phase of the build. Measured cold compile of a single GEMV (shape M=448, K=64, columns=8, build only, no device run):

| num_batches | unrolled | coalesced |
|---|---|---|
| 192 | 5.99s | 4.72s |
| 1536 | 52.37s | 3.54s |

The coalesced build is roughly constant in `num_batches` while the unrolled build grows linearly, so the saving scales with the batch count.

## Validation
- Offline access-equivalence of the coalesced vs unrolled descriptor for the tested and fallback configurations, including the alignment-aware split.
- `num_batches == 1` generates byte-identical lowered output to the previous code.
- The new device tests pass on NPU2 at `num_batches` of 4, 8, 32, 100, and 192, plus the aligned-split shape (M=1026), the fallback shape (batch stride > limit), and an attention-style shape (`tile_size_input=4`); the existing `num_batches == 1` tests are unchanged.
- Exercised end to end: the in-repo `llama_3.2_1b` application, which calls GEMV with `num_batches = n_heads`, builds and generates coherent output on NPU2 with the coalesced path.
- In-repo GEMV consumers were checked: `swiglu_decode` uses `num_batches=1` (byte-identical, unaffected); the `llama_3.2_1b` application calls GEMV with `num_batches = n_heads` for the attention scores/context GEMVs - those now coalesce, and the coalesced descriptor is legal and access-equivalent at those shapes, so behavior is unchanged.

## PR Merge Checklist

### DIFF
--- a/iron/operators/gemv/design.py
+++ b/iron/operators/gemv/design.py
@@ -165,6 +165,70 @@ def my_matvec(
         for col in range(cols)
     ]
 
+    # --- Batch-coalesce (default): one BD per column over all batches. ---
+    # Replaces the per-batch unroll with a single iterated BD; the stock A_taps/C_taps
+    # above remain the fallback (and the num_batches==1 path). Access-equivalent to the
+    # unroll (covered by test_gemv_batched).
+    # AIE shim BD limits (mlir-aie AIEXDialect.cpp verifyStridesWraps): the two wrap
+    # dims have a size cap (1023) while one dim is size-uncapped; TAP sizes are
+    # outermost-first and the verifier reverses them, so [1, num_batches, run_hi,
+    # run_lo] places num_batches in the uncapped dim and the contiguous run in the two
+    # wrap dims. Access-equivalent to the unroll (covered by test_gemv_batched).
+    # The shim also enforces a 4-byte address granularity on every transfer size and
+    # stride (NOT skipped, even for linear transfers); for bf16 (2 bytes) that means
+    # the innermost size (run_lo) and the batch stride must be even, so split_run only
+    # yields an even run_lo and the predicate requires even strides.
+    MAX_WRAP = 1023
+    MAX_STRIDE = (1 << 20) - 1  # conservative element-stride bound for the wrap dims
+    GRAN_ELEMS = 2  # 4-byte shim granularity / 2-byte bf16 element
+
+    def split_run(run, lim=MAX_WRAP, gran=GRAN_ELEMS):
+        """Factor a contiguous run into (hi, lo), both <= lim and lo a multiple of gran
+        (the address-granularity-aligned inner size), lo maximal. None if no such
+        split exists (caller then falls back to the per-batch path)."""
+        lo_start = (lim // gran) * gran
+        for lo in range(lo_start, 0, -gran):
+            if run % lo == 0 and (run // lo) <= lim:
+                return (run // lo, lo)
+        return None
+
+    A_run, A_bstride = (M // cols) * K, M * K
+    C_run, C_bstride = (M // cols), M
+    A_split, C_split = split_run(A_run), split_run(C_run)
+    coalesce = (
+        num_batches > 1
+        and A_bstride <= MAX_STRIDE
+        and C_bstride <= MAX_STRIDE
+        and A_bstride % GRAN_ELEMS == 0
+        and C_bstride % GRAN_ELEMS == 0
+        and A_split is not None
+        and C_split is not None
+    )
+
+    def coalesced_tap(L3_ty, col_off, split, bstride):
+        run_hi, run_lo = split
+        return TensorAccessPattern(
+            tensor_dims=L3_ty.__args__[0],
+            offset=col_off,
+            sizes=[1, num_batches, run_hi, run_lo],
+            strides=[0, bstride, run_lo, 1],
+        )
+
+    if coalesce:
+        # Backpressure replaces the per-batch drain wait, so the A/C ObjectFifos must
+        # be deep enough (>=2) for the producer not to overrun the consumer.
+        assert all(f.depth >= 2 for f in A_L3L1_fifos) and all(
+            f.depth >= 2 for f in C_L1L3_fifos
+        ), "coalesced GEMV needs A/C ObjectFifo depth>=2 (replaces the per-batch wait)"
+        A_taps_coalesced = [
+            coalesced_tap(L3_A_ty, col * (M // cols) * K, A_split, A_bstride)
+            for col in range(cols)
+        ]
+        C_taps_coalesced = [
+            coalesced_tap(L3_C_ty, col * (M // cols), C_split, C_bstride)
+            for col in range(cols)
+        ]
+
     rt = Runtime()
     with rt.sequence(L3_A_ty, L3_B_ty, L3_C_ty) as (A, B, C):
         rt.start(*workers)
@@ -172,21 +236,43 @@ def my_matvec(
         for col in range(cols):
             # Simple linear transfer of B, includes all batches in sequence
             rt.fill(B_L3L1_fifos[col].prod(), B, B_tap, task_group=tg_b)
-        for batch in range(num_batches):
+        if coalesce:
+            # One iterated BD per column over all batches; per-batch `wait` dropped
+            # in favor of ObjectFifo backpressure (asserted above).
             tg_ac = rt.task_group()
             for col in range(cols):
                 rt.fill(
-                    A_L3L1_fifos[col].prod(), A, A_taps[col][batch], task_group=tg_ac
+                    A_L3L1_fifos[col].prod(), A, A_taps_coalesced[col], task_group=tg_ac
                 )
             for col in range(cols):
                 rt.drain(
                     C_L1L3_fifos[col].cons(),
                     C,
-                    C_taps[col][batch],
+                    C_taps_coalesced[col],
                     task_group=tg_ac,
                     wait=True,
                 )
             rt.finish_task_group(tg_ac)
+        else:
+            # Fallback (also the num_batches==1 path): stock per-batch unroll.
+            for batch in range(num_batches):
+                tg_ac = rt.task_group()
+                for col in range(cols):
+                    rt.fill(
+                        A_L3L1_fifos[col].prod(),
+                        A,
+                        A_taps[col][batch],
+                        task_group=tg_ac,
+                    )
+                for col in range(cols):
+                    rt.drain(
+                        C_L1L3_fifos[col].cons(),
+                        C,
+                        C_taps[col][batch],
+                        task_group=tg_ac,
+                        wait=True,
+                    )
+                rt.finish_task_group(tg_ac)
         rt.finish_task_group(tg_b)
 
     return Program(dev, rt).resolve_program()
--- a/iron/operators/gemv/design.py
+++ b/iron/operators/gemv/design.py
@@ -165,34 +165,15 @@ def my_matvec(
         for col in range(cols)
     ]
 
-    # --- Batch-coalesce (default): one BD per column over all batches. ---
-    # Replaces the per-batch unroll with a single iterated BD; the stock A_taps/C_taps
-    # above remain the fallback (and the num_batches==1 path). Access-equivalent to the
-    # unroll (covered by test_gemv_batched).
+    # Batch coalescing replaces the per-batch unroll with a single iterated BD.
     #
-    # This is NOT a single linear transfer. Within one batch the run is contiguous
-    # (A_run = (M//cols)*K elements), but the batch stride is the full matrix
-    # (A_bstride = M*K), so for cols>1 each column gathers its own slice out of every
-    # batch with a gap in between. Only cols==1 degenerates to bstride==run. So the
-    # batch dim is a genuine size-uncapped iteration dim and the TAP is required.
+    # Within one batch the run is contiguous (A_run = (M//cols)*K elements).
+    # The batch stride is the full matrix (A_bstride = M*K), so for cols>1 each column 
+    # gathers its own slice out of every batch with a gap in between. 
     #
     # The contiguous run is then split into two wrap dims [run_hi, run_lo] ONLY to fit
-    # the AIE shim's 10-bit (1023) wrap-size cap. TAP sizes are outermost-first and the
-    # verifier reverses them, so [1, num_batches, run_hi, run_lo] puts num_batches in the
-    # size-uncapped dim and the contiguous run in the two capped wrap dims. The shim also
-    # enforces a 4-byte address granularity on every size and stride (not skipped, even
-    # for linear transfers); for bf16 (2 bytes) that means run_lo and the batch stride
-    # must be even, so split_run only yields an even run_lo and the predicate requires
-    # even strides.
+    # the AIE shim's 10-bit (1023) wrap-size cap.
     #
-    # NOTE: the split cannot be dropped in the cols>1 case, even on a pin with the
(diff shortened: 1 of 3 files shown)


iron/operators/gemv/design.py
```diff
@@ -166,28 +166,114 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
+        return TensorAccessPattern(
+            tensor_dims=L3_ty.__args__[0],
+            offset=col_off,
+            sizes=[1, num_batches, run_hi, run_lo],
+            strides=[0, bstride, run_lo, 1],
+        )
+
+    if coalesce:
+        # Backpressure replaces the per-batch drain wait, so the A/C ObjectFifos must
+        # be deep enough (>=2) for the producer not to overrun the consumer.
```
andrej: What do you mean by the producer overrunning the consumer? Proper lock synchronization and backpressure should i.m.o. lead to at most a deadlock, and "overrunning" (buffers?) sounds concerning.

iron/operators/gemv/design.py
```diff
@@ -166,28 +166,114 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
 
+    # --- Batch-coalesce (default): one BD per column over all batches. ---
+    # Replaces the per-batch unroll with a single iterated BD; the stock A_taps/C_taps
+    # above remain the fallback (and the num_batches==1 path). Access-equivalent to the
+    # unroll (covered by test_gemv_batched).
+    # AIE shim BD limits (mlir-aie AIEXDialect.cpp verifyStridesWraps): the two wrap
+    # dims have a size cap (1023) while one dim is size-uncapped; TAP sizes are
+    # outermost-first and the verifier reverses them, so [1, num_batches, run_hi,
+    # run_lo] places num_batches in the uncapped dim and the contiguous run in the two
+    # wrap dims. Access-equivalent to the unroll (covered by test_gemv_batched).
```
andrej:
```suggestion
    # wrap dims.
```

iron/operators/gemv/design.py
```diff
@@ -166,28 +166,114 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
+
+    def split_run(run, lim=MAX_WRAP, gran=GRAN_ELEMS):
+        """Factor a contiguous run into (hi, lo), both <= lim and lo a multiple of gran
+        (the address-granularity-aligned inner size), lo maximal. None if no such
+        split exists (caller then falls back to the per-batch path)."""
+        lo_start = (lim // gran) * gran
+        for lo in range(lo_start, 0, -gran):
+            if run % lo == 0 and (run // lo) <= lim:
+                return (run // lo, lo)
+        return None
```
andrej, line 192: @atassis Does this split up a purely linear, contiguous transfer? There is a special case to _not_ use data layout transformation dimensions at all. If left off, you should be able to specify only transfer lengths, and no dimensions. 

@hunhoffe My understanding is this splits up a contiguous transfer into multiple dimensions. Is this logic still required after MLIR-AIE PRs [#2924](https://github.com/Xilinx/mlir-aie/pull/2924) and [#3036](https://github.com/Xilinx/mlir-aie/pull/3036)?

iron/operators/gemv/design.py
```diff
@@ -166,28 +166,114 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
+    # outermost-first and the verifier reverses them, so [1, num_batches, run_hi,
+    # run_lo] places num_batches in the uncapped dim and the contiguous run in the two
+    # wrap dims. Access-equivalent to the unroll (covered by test_gemv_batched).
+    # The shim also enforces a 4-byte address granularity on every transfer size and
+    # stride (NOT skipped, even for linear transfers); for bf16 (2 bytes) that means
+    # the innermost size (run_lo) and the batch stride must be even, so split_run only
+    # yields an even run_lo and the predicate requires even strides.
+    MAX_WRAP = 1023
+    MAX_STRIDE = (1 << 20) - 1  # conservative element-stride bound for the wrap dims
+    GRAN_ELEMS = 2  # 4-byte shim granularity / 2-byte bf16 element
```
andrej, line 182:
```suggestion
    MAX_WRAP = 1023
    MAX_STRIDE = (1 << 20) - 1  # conservative element-stride bound for the wrap dims
    GRAN_ELEMS = 2  # 4-byte shim granularity / 2-byte bf16 element
    # FIXME: Pull this information in to the target model in MLIR-AIE; also used in https://github.com/Xilinx/mlir-aie/blob/58ad59e8b3a601f540a653d1222d1dfc6303d3b0/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L191-L199
```

(No need to do in this PR but let's leave the comment as a future to do)

[Reference](https://github.com/Xilinx/mlir-aie/blob/58ad59e8b3a601f540a653d1222d1dfc6303d3b0/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L191-L199)

iron/operators/gemv/design.py
```diff
@@ -166,28 +166,114 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
+    )
+
+    def coalesced_tap(L3_ty, col_off, split, bstride):
+        run_hi, run_lo = split
+        return TensorAccessPattern(
+            tensor_dims=L3_ty.__args__[0],
+            offset=col_off,
+            sizes=[1, num_batches, run_hi, run_lo],
+            strides=[0, bstride, run_lo, 1],
+        )
```
andrej, line 214: I didn't spend too much time thinking through this but at a glance this looks like it might just be a linear transfer (if bstride==run_hi*run_lo). If that's the case, a lot of this code can be deleted and you should be able to just _not_ supply any TAP to the `rt.fill` at all, which will equal a linear transfer.

If the transfer is more complicated, please explain in a comment! Thanks

iron/operators/gemv/design.py
```diff
@@ -166,28 +166,114 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
+                    )
+                for col in range(cols):
+                    rt.drain(
+                        C_L1L3_fifos[col].cons(),
+                        C,
+                        C_taps[col][batch],
+                        task_group=tg_ac,
+                        wait=True,
+                    )
+                rt.finish_task_group(tg_ac)
```
andrej: Since this is essentially identical to the if-branch above except for number of wait/finish_task_group calls, could we reduce code duplication by instead looping over a new variable `num_waits`, which in this else-case would be `num_waits==num_batches` and in the if-case would be `num_waits==1`, and then also swap out the TAPs as appropriate.

andrej (changes requested):
Thanks, looking good! Reducing the number of runtime sequence instructions seems useful and the performance gain seems significant.

I only have some questions about the tensor access patterns. I have a feeling / hope that this could be radically simplified if indeed they are just linear transfers.

andrej:
Oh, one more thing, could you please add the latency/gflops metrics to the batched tests as well? I know it's not technically part of this PR but it would be nice to keep track of the performance numbers in the CI results:

```    gflops = (2.0 * M * K) / (latency_us * 1e-6) / 1e9
    print(f"Throughput: {gflops:.6e} GFLOP/s")
    print(f"Effective Bandwidth: {bandwidth_gbps:.6e} GB/s\n")
```

image

image

iron/operators/gemv/design.py
atassis: So within a batch the run is contiguous, but the batch stride is the whole matrix (M*K) and each column only takes its (M//cols)*K slice, so for cols>1 the batches have gaps between them - it's linear only when cols==1. On #2924/#3036: yeah those would let me drop the inner split, but we pin mlir_aie e4f35d6 (Mar 31) and both landed after that (Apr 7 / May 11), so on this pin the split is still needed. I left a note that it can become [num_batches, A_run] once the pin moves past #3036.

iron/operators/gemv/design.py
atassis: Yeah I checked that - bstride is M*K and the run is (M//cols)*K, so they're equal only when cols==1. For cols>1 it's a real strided gather, that's why the batch dim has to stay. The run_hi/run_lo split is just to get under the 1023 wrap cap (more in the other thread). So I explained it in a comment instead of deleting.

iron/operators/gemv/design.py
atassis: Yeah, "overrun" is a bad word for it, sorry. The ObjectFifo lock backpressures, so a fast producer just blocks (worst case it stalls), it can't corrupt anything. depth>=2 is only there to overlap the fill with compute, so it's a perf thing, not correctness. Reworded the comment.

iron/operators/gemv/design.py
atassis: Added the FIXME pointing at verifyStridesWraps in AIEXDialect.cpp. Kept the constants, still needed on this pin.

iron/operators/gemv/design.py
atassis: Done. Also dropped the duplicated sentence.

iron/operators/gemv/design.py
atassis: Done - one loop over num_waits now (1 for coalesced, num_batches otherwise), and swap the tap.

atassis:
Pushed all of the above. I also added `@pytest.mark.metrics` to `test_gemv_batched` - it was correctness-only before, so the coalescing speedup wasn't showing up in CI.

iron/operators/gemv/design.py
andrej: @atassis Just updated the pinned MLIR-AIE. Can that simplify this?

iron/operators/gemv/design.py
andrej: Awesome! thanks, I prefer that

andrej:
Thanks for looking into those suggestions. I did just update the pinned MLIR-AIE. If you can look into whether that helps you simplify any of the linearizing code, that would be great. After that, or if it doesn't end up helping you, I'll run the workflows and merge :)

atassis:
@andrej Hi there
I have  looked into it on v1.3.4. The contiguous-run linearization (#2924/#3036) only drops the split when the whole transfer is contiguous, i.e. cols==1 (bstride==run) - there [1, num_batches, run] lowers to a single buffer-length transfer. For a real cols>1 gather the batch stride leaves a gap, so the inner run stays a wrap dim: [1, num_batches, 4096] fails aiecc with "Size 4096 exceeds [0:1023]". So the split stays for cols>1. I rebased onto devel and migrated to the new placer; the full gemv suite (80 tests, batched + non-batched) passes on NPU2 with the v1.3.4 wheel. Ready for your workflows whenever

iron/operators/gemv/design.py
```diff
@@ -165,24 +165,115 @@ def core_body(A_L3L1_fifo, B_L3L1_fifo, C_L1L3_fifo, matvec):
...
+    # contiguous-run linearization (mlir-aie #2924 + #3036, present in v1.3.4). That
+    # linearization only bypasses the 1023 wrap cap when the WHOLE transfer is
+    # contiguous, i.e. bstride==run (cols==1); there [1, num_batches, run] with strides
+    # [0, bstride, 1] lowers to one linear buffer-length transfer. For cols>1 the batch
+    # stride leaves a gap, so the inner run stays a genuine wrap dim: verified on v1.3.4
+    # that [1, num_batches, 4096] (a cols=8 run) fails aiecc with "Size 4096 exceeds the
+    # [0:1023] range", while the split below lowers and runs. So the split stays.
+    # FIXME: pull these shim BD bounds from the MLIR-AIE target model rather than
+    # hard-coding them; they live in verifyStridesWraps in
+    # https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/IR/AIEXDialect.cpp
```
andrej:
```suggestion
    # Batch coalescing replaces the per-batch unroll with a single iterated BD.
    #
    # Within one batch the run is contiguous (A_run = (M//cols)*K elements).
    # The batch stride is the full matrix (A_bstride = M*K), so for cols>1 each column 
    # gathers its own slice out of every batch with a gap in between. 
    #
    # The contiguous run is then split into two wrap dims [run_hi, run_lo] ONLY to fit
    # the AIE shim's 10-bit (1023) wrap-size cap.
    #
    # FIXME: pull these shim BD bounds from the MLIR-AIE target model rather than
    # hard-coding them; they live in verifyStridesWraps in
    # https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/IR/AIEXDialect.cpp
```

andrej (approved):

andrej:
@atassis Awesome, thank you for looking into the linearization even if it didn't apply here. The other changes cleaned up the code quite nicely. I will merge this as soon as tests pass. Remind me if I forget please!

atassis:
@andrej No issue! I'll just remind then there are 2 more PRs in IRON from me, if I am not mistaken. #134 #135 
I can rebase them onto latest upstream if needed

PR #131, opened 2026-07-06
Use ScratchpadParameter feature in Llama

andrej:
Replace the ugly magic-value patching in the llama example application for run-time parameters with the new ScratchpadParameter feature in MLIR-AIE.

Builds on #117 -- [isolated diff](https://github.com/andrej/IRON/compare/fused-dispatch-modes...scratchpad)

### DIFF
--- a/iron/common/compilation/base.py
+++ b/iron/common/compilation/base.py
@@ -286,21 +286,23 @@ class SourceArtifact(CompilationArtifact):
     pass
 
 
+class MLIRArtifact(CompilationArtifact):
+    """Base class for artifacts whose file is an MLIR (.mlir) module usable as aiecc input.
+
+    ``_MLIRInputMixin.mlir_input`` locates the MLIR source of a downstream
+    target (elf/xclbin/insts.bin) by looking for a dependency of this type.
+    Using a shared base class (rather than name-checking) lets other modules
+    such as ``compilation/sequence.py`` opt in without creating an import cycle.
+    """
+
+
 class _MLIRInputMixin:
     """Mixin providing a mlir_input property that finds the MLIR source in dependencies."""
 
     @property
     def mlir_input(self):
-        # Accept any MLIR-source-style artifact in dependencies. We name-check
-        # ``SequenceMLIRSource`` to avoid an import cycle with
-        # ``compilation/sequence.py`` (which itself imports from this module).
         result = next(
-            (
-                d
-                for d in self.dependencies
-                if isinstance(d, (SourceArtifact, PythonGeneratedMLIRArtifact))
-                or type(d).__name__ == "SequenceMLIRSource"
-            ),
+            (d for d in self.dependencies if isinstance(d, MLIRArtifact)),
             None,
         )
         if result is None:
@@ -375,7 +377,7 @@ class KernelArchiveArtifact(CompilationArtifact):
     pass
 
 
-class PythonGeneratedMLIRArtifact(CompilationArtifact):
+class PythonGeneratedMLIRArtifact(MLIRArtifact):
     def __init__(
         self,
         filename: str,
@@ -471,11 +473,9 @@ class GenerateMLIRFromPythonCompilationRule(CompilationRule):
         commands = []
         worklist = graph.get_worklist(PythonGeneratedMLIRArtifact)
         for artifact in worklist:
-            new_artifact = SourceArtifact(artifact.filename)
-            callback = partial(self.generate_mlir, new_artifact, artifact.generator)
+            callback = partial(self.generate_mlir, artifact, artifact.generator)
             commands.append(PythonCallbackCompilationCommand(callback))
-            new_artifact.available = True
-            graph.replace(artifact, new_artifact)
+            artifact.available = True
         return commands
 
     @staticmethod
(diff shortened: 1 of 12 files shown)


jgmelber (approved):

iron/common/compilation/base.py
```diff
@@ -291,11 +291,15 @@ class _MLIRInputMixin:
...
     def mlir_input(self):
+        # Accept any MLIR-source-style artifact in dependencies. We name-check
+        # ``SequenceMLIRSource`` to avoid an import cycle with
+        # ``compilation/sequence.py`` (which itself imports from this module).
         result = next(
             (
                 d
                 for d in self.dependencies
                 if isinstance(d, (SourceArtifact, PythonGeneratedMLIRArtifact))
+                or type(d).__name__ == "SequenceMLIRSource"
```
hunhoffe: Why is checking the name of the type necessary? Seems a bit low-level...

iron/common/compilation/sequence.py
```diff
@@ -90,20 +90,34 @@ def get_child_mlir_module(mlir_artifact: PythonGeneratedMLIRArtifact) -> Any:
...
         mlir_module = get_child_mlir_module(mlir_artifact)
-        device_ops = [
-            op for op in mlir_module.body.operations if isinstance(op, aie.DeviceOp)
-        ]
+        device_ops = []
+        params_here: dict[str, ir.Type] = {}
+        for op in mlir_module.body.operations:
+            if isinstance(op, aie.DeviceOp):
+                device_ops.append(op)
+            elif op.operation.name == "aiex.scratchpad_parameter":
```
hunhoffe, line 116: I think this is OK for now, but long term -- to keep the code base clean -- I really try to avoid doing actual operation/MLIR parsing in Python and leave that to the compiler.

iron/common/compilation/sequence.py
```diff
@@ -113,14 +127,47 @@ def fuse_mlir(artifact: FusedMLIRSource) -> None:
         if device_ty is None:
             device_ty = device_op.device
         device_mlir_strings[op_name] = str(device_op)
+        operator_param_decls[op_name] = params_here
         sequence_arg_types[op_name] = extract_runtime_sequence_arg_types(device_op)
 
+    # Deduplicate parameter decls across operators (same name must have the
+    # same type; otherwise indices would collide in the global state table).
+    hoisted_params: dict[str, ir.Type] = {}
```
hunhoffe, line 134: This definitely seems like it should be a check in the compiler?

iron/common/sequence.py
```diff
@@ -0,0 +1,814 @@
...
+        # The last xclbin in the chain carries all the linked instances.
+        artifacts.append(prev_xclbin)
+        self.combined_xclbin = prev_xclbin
+        seq.add_artifacts(artifacts)
+
+    def make_callable(self, seq):
+        return SequenceXclbinCallable(seq, self)
+
+
+class CompareDispatch(SeparateDispatch):
```
hunhoffe, line 188: Do you think -- maybe in another PR -- we could try to converge on some utilities?
* https://github.com/Xilinx/mlir-aie/blob/main/python/utils/benchmark.py
* https://github.com/Xilinx/mlir-aie/blob/main/python/utils/verify.py
* https://github.com/Xilinx/mlir-aie/blob/main/python/utils/test.py

I believe pretty much all the programming examples in mlir-aie have been ported to use common utilities; would be create to converge on a shared solution if possible.

iron/common/utils.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import numpy as np
+import torch
```
hunhoffe: Is torch always a required import? I can't remember.

iron/tests/infrastructure/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+
+from iron.common.sequence import OperatorSequence
+from iron.common.compilation.sequence import fuse_mlir
+from iron.common.test_utils import verify_buffer
+from iron.operators.elementwise_add.op import ElementwiseAdd
+from iron.operators.relu.op import ReLU
+
+
+def _is_npu2():
+    return isinstance(aie_utils.get_current_device(), NPU2)
```
hunhoffe: Seems a weird utility function to put here. Are you sure there isn't a shared utility function in mlir-aie that does this? Or should this be refactored out into a utility function?

hunhoffe (approved):
Nice work! Cool to see the ScratchpadParameter at work.

iron/common/compilation/sequence.py
andrej: discussed offline -- future work due to operator metadata currently living at IRON-level and doing this at MLIR-AIE-level would require migrating this metadata infrastructure

iron/common/compilation/sequence.py
andrej: discussed offline -- future work due to operator metadata currently living at IRON-level and doing this at MLIR-AIE-level would require migrating this metadata infrastructure

iron/common/sequence.py
andrej: discussed offline -- future work

PR #132, opened 2026-07-09
Make SwiGLU operator use the `OperatorSequence` abstraction

andrej:
Remove custom `--xclbin-input` logic in the layer-by-layer SwiGLU operator in favor of the same `OperatorSequence` (ex. `FusedMLIROperator`) abstraction that the llama application uses. Makes for a better, more minimal use case example of `OperatorSequence` to point people to. Should help with performance, since this will use the single-dispatch full-ELF `aiex.configure` flow on Strix.


andrej:
image
CI performance regreessions confirm another example of the `--expand-load-pdis` speedups :)

hunhoffe (approved):
Nice simplification!

jgmelber (approved):

PR #134, opened 2026-07-10
compilation: honor AIECC_PATH and AIECC_JOBS env overrides

atassis:
Two opt-in, backward-compatible build-ergonomics knobs on the aiecc compilation rules. Both default to the current behavior, so nothing changes unless the env var is set.

- **`AIECC_PATH`** - point the compilation rules at a locally-built `aiecc` (for example a compiler under active development) instead of the one under `mlir_aie_dir/bin`, without having to overwrite the installed binary. Default: the installed `aiecc`.
- **`AIECC_JOBS`** - pass `-j<N>` to `aiecc` instead of the hardcoded `-j1`, so multi-core hosts can compile the AIE cores in parallel. Default `1` = unchanged.

Both are read via `os.environ.get(...)` in `iron/common/compilation/base.py` (the `aiecc` path resolution and the two `Aiecc*CompilationRule` command builders). No new dependencies.

### DIFF
--- a/iron/common/compilation/base.py
+++ b/iron/common/compilation/base.py
@@ -490,7 +490,14 @@ class AieccCompilationRule(CompilationRule):
         self, build_dir, peano_dir, mlir_aie_dir, use_chess=False, *args, **kwargs
     ):
         self.build_dir = build_dir
-        self.aiecc_path = Path(mlir_aie_dir) / "bin" / "aiecc"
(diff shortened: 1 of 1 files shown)


andrej (approved):

atassis:
Hi there, @andrej ! Friendly ping if this is ok to merge.

PR #135, opened 2026-07-10
gemv: add optional fused GELU epilogue (NPU2)

atassis:
Adds an opt-in `epilogue` parameter to `GEMV`. `epilogue="none"` (default) is byte-identical to today; `epilogue="gelu"` fuses a GELU (tanh approximation) over each output tile inside the producing core, folding an activation that would otherwise be a separate elementwise pass over the GEMV output.

### Details
- The GELU runs once per full `m_output` tile in `core_body` (after the matvec inner-loop has filled all rows), not per matvec call whose `m_input` tile can be smaller than the 16-wide activation vector.
- **No duplicated math.** `aie_kernels/aie2p/gelu.cc` is refactored so the per-16-lane computation is a shared helper used by both the existing out-of-place `gelu_tanh_approx_bf16` and a new in-place `gelu_tile_bf16`. In-place is aliasing-correct (a single `restrict` pointer, each slot read then written); reusing the out-of-place kernel with `input == output` would violate its `restrict` contract, so a dedicated in-place entry point is the correct way to share the code.
- Since `gelu.cc` lives in `aie2p/`, the fused epilogue is **NPU2-only**; `GEMV` raises `NotImplementedError` if it is requested on NPU1. When enabled, the core links a `(matvec, gelu)` archive; the default path still links the single matvec object.

### Test
`test_gemv_gelu` compares the fused output against a `gelu(A @ B)` golden across three shapes, using the same latency/bandwidth/throughput `@pytest.mark.metrics` as `test_gemv`. **Verified on NPU2: 15/15 pass.**

This is opt-in and default-preserving; it also composes with the existing `func_prefix` fusion support (#123) for use as a fused block.

### DIFF
--- a/aie_kernels/aie2p/gelu.cc
+++ b/aie_kernels/aie2p/gelu.cc
@@ -8,19 +8,12 @@
 
 using namespace aie;
 
-void gelu_tanh_approx_bf16(bfloat16 *restrict input_vector, bfloat16 *restrict output_vector, const int32_t vector_size)
+// One 16-lane GELU (tanh approximation): 0.5 * x * (1 + tanh(sqrt(2/pi) * (x + 0.044715 * x^3))).
+static inline aie::vector<bfloat16, 16> gelu_tanh_approx_v16(aie::vector<bfloat16, 16> x)
 {
-    event0();
-
-    auto it_in = aie::begin_restrict_vector<16>((bfloat16 *)input_vector);
-    auto it_out = aie::begin_restrict_vector<16>((bfloat16 *)output_vector);
-
-    aie::vector<bfloat16, 16> input;
-
-    // Constants
     const bfloat16 k0_5 = 0.5f;
     const bfloat16 k1 = 1.0f;
-    const bfloat16 sqrt_2_over_pi = 0.79788456f; // ≈ sqrt(2/π)
+    const bfloat16 sqrt_2_over_pi = 0.79788456f; // sqrt(2/pi)
     const bfloat16 kBeta = 0.044715f;
 
     auto v05 = aie::broadcast<bfloat16, 16>(k0_5);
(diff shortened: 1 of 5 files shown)


andrej (approved):
This is a good addition and a natural fusion opportunity. If you have any interest in this, I think it would be cool to think about if we can generalize this across other activation types as well -- sort of a generic "run this kernel in-place on the results" flag, where maybe you could pass an arbitrary `Kernel` object. That would be a follow-on PR of course.

atassis:
I am already sticking to a generic "run an arbitrary Kernel in-place on the output" epilogue slot, not a one-off gelu branch.

The genericity matters to me because I'm building a full NPU engine backend on XDNA2, so composing generic primitives instead of hand-fusing per model is the design rule I hold to - a generic epilogue slot fits that squarely. The harder piece I'm on right now is the resident fused dataflow: keeping the whole transformer block on-chip across dispatches, with on-chip KV and decode attention, to cut out the host round-trips - and that only holds together if the primitives stay composable. If you have a take on whether that's the right direction to push, I'd genuinely value it; you all have far more context on where the toolchain is heading.

On the PR itself: I'll rebase onto the latest devel (the conflict is just from my batched-gemv PR that merged) and clear the lint and any other CI so it lands clean. Happy to do the generic epilogue as a follow-on once this lands.

atassis:
@andrej Hi there. I have fixed a CI related issue and another latent bug

atassis:
@andrej on generalizing - here's the shape I'd put up: an optional in-place epilogue on the gemv output tile. You pass the elementwise kernel (its symbol, via the func_prefix hook from #123) and it runs on the tile before write-back, so GELU becomes one instance instead of a dedicated branch, same single-dispatch fusion.

Two things I'd want your read on: scope - keep it gemv-only to start, or make the epilogue slot generic across the producer ops (gemm etc.) from the start? And whether func_prefix is the right way to name the epilogue kernel, or you'd rather a small activation registry.

I'll keep this PR as the GELU case so it can land, and put the generalization up as a follow-up.

atassis:
Hi, @andrej!
Rebased on devel, CI green. I'd like to land this as-is. I'll do the generic in-place-kernel epilogue you suggested as the follow-on. OK to merge?

andrej:
@atassis Yes, sorry for the delay -- will merge this now

andrej:
@atassis Regarding generalization (you of course do not have to take this on, but here are my thoughts):
> Two things I'd want your read on: scope - keep it gemv-only to start, or make the epilogue slot generic across the producer ops (gemm etc.) from the start? And whether func_prefix is the right way to name the epilogue kernel, or you'd rather a small activation registry.

GEMV-only to start would be fine.

Rather than an activation registry I'd like to see it support any `ChanneledUnaryOperator`. Pass in the operator class itself, and the suffix extracts the kernel from `get_kernel_artifacts`. The `ChanneledUnaryOperator` already exposes `kernel_fn_name`.

PR #139, opened 2026-07-19
rms_norm: fix bf16 accuracy (f32 normalize + conv_even) and parameterize epsilon

atassis:
## Summary

Two related changes to the aie2/aie2p bf16 RMSNorm kernel and its IRON operator:

1. An accuracy fix: normalize in f32 and round once, and set the rounding mode to
   round-to-nearest-even (`conv_even`).
2. Make `epsilon` a parameter instead of a hardcoded `1e-5f`.

The default behavior is unchanged (epsilon still defaults to 1e-5). Callers opt
into a different value with `RMSNorm(..., epsilon=...)`.

## Motivation

I hit this bringing a small LLM decode (Gemma-3-270M, 18 fused layers) up on
aie2p. The generated tokens matched a host reference for the first few decode
steps and then flipped on a step where two candidate tokens were close in logit
space. The graph itself was correct: an f32 (and a faithful round-to-nearest
bf16) reference of the same graph reproduced the reference output. The device was
noisier than a correct bf16 forward should be, and the error grew roughly
linearly with depth rather than as sqrt(depth).

Two things in the RMSNorm kernel caused that:

- `inv_rms` (the reciprocal RMS scalar) was cast to bf16 and then multiplied
  bf16*bf16, so the entire normalized vector was scaled by one imprecise scalar
  per norm. That is a coherent, same-direction error per norm. Over a deep
  residual stream it accumulates roughly linearly, which is what made it large
  enough to flip a near-tie argmax; unbiased round-to-nearest error would grow
  only as sqrt(N) and partly cancel.
- The kernel set no rounding mode, so it ran under whatever a previously
  dispatched kernel left in the core's global rounding state. `set_rounding` is
  global per core, so in a fused graph a kernel that leaves floor/truncation set
  poisons the ones that follow.

Separately, `epsilon` was hardcoded to `1e-5f` (Llama's value). Gemma uses 1e-6,
and it is not a rounding detail there: Gemma's deep post-feedforward sandwich
norms operate on near-zero inputs where `mean(x^2)` is comparable to epsilon, so
epsilon sets the normalization scale.

## What changed

Commit 1, "normalize in f32 and set conv_even rounding":
- `aie_kernels/{aie2,aie2p}/rms_norm.cc`: normalize through an `accfloat`
  accumulator (mirroring `layer_norm.cc`) and round once at the output; call
  `set_rounding(conv_even)` at both entry points.

Commit 2, "parameterize epsilon":
- `aie_kernels/{aie2,aie2p}/rms_norm.cc`: `rms_norm_general` /
  `rms_norm_bf16_vector` / `weighted_rms_norm` take a `float epsilon` argument.
- `iron/operators/rms_norm/{design,design_weighted}.py`: declare the `np.float32`
  kernel argument and pass `epsilon` (a per-instance compile-time constant, the
  same mechanism the axpy op uses for its scalar factor).
- `iron/operators/rms_norm/op.py`: add an `epsilon` field (default 1e-5).
- `iron/operators/rms_norm/reference.py`: accept `eps` and align the formula to
  the kernel, `1/sqrt(mean(x^2) + eps)` rather than `1/(sqrt(mean(x^2)) + eps)`.

## Testing

On aie2p (Strix):
- Isolated weighted RMSNorm rel-L2 vs a host reference improved from ~0.008 to
  ~0.0025.
- A fused 18-layer Gemma-3-270M decode that was previously flipping a token now
  matches its host reference exactly (8 of 8 greedy tokens), deterministic across
  repeated runs, with the op passing epsilon=1e-6.

The aie2 (Phoenix) kernel gets the same source change and compiles; I do not have
aie2 hardware, so it is compile-verified, not silicon-verified. The change mirrors
the aie2p one and the existing aie2 `layer_norm.cc` accfloat path.

### DIFF
--- a/aie_kernels/aie2/rms_norm.cc
+++ b/aie_kernels/aie2/rms_norm.cc
@@ -34,9 +34,7 @@ void rms_norm_general(const T *restrict input, const T *restrict input2, T *rest
 
     float rms = sum_sq / cols + epsilon;
     float inv_rms = invsqrt(rms);
-    // Normalize in f32 and round once. The previous code cast inv_rms to bf16 and did a
-    // bf16*bf16 multiply, a coherent per-norm scale error that accumulated over a deep
-    // residual stream and flipped near-tie argmaxes. Mirrors layer_norm.cc's accfloat path.
+    // Normalize in f32 and round once/ Mirrors layer_norm.cc's accfloat path.
     ::aie::accum<accfloat, N> inv_rms_v;
     inv_rms_v.from_vector(::aie::broadcast<float, N>(inv_rms), 0);
 
@@ -71,13 +69,13 @@ void rms_norm_general(const T *restrict input, const T *restrict input2, T *rest
 extern "C" {
 void rms_norm_bf16_vector(bfloat16 *input, bfloat16 *output, int32_t size, float epsilon)
 {
-    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even; do not inherit a floor rounding mode from a prior kernel
+    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even
     rms_norm_general<bfloat16, 16>(input, nullptr, output, size, epsilon);
 }
 
 void weighted_rms_norm(bfloat16 *a_in, bfloat16 *b_in, bfloat16 *c_out, int32_t size, float epsilon)
 {
-    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even; do not inherit a floor rounding mode from a prior kernel
+    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even
     rms_norm_general<bfloat16, 16>(a_in, b_in, c_out, size, epsilon);
 }
 }
--- a/aie_kernels/aie2/rms_norm.cc
+++ b/aie_kernels/aie2/rms_norm.cc
@@ -9,7 +9,11 @@
 #include <stdlib.h>
 
 template <typename T, int N>
-void rms_norm_general(const T *restrict input, const T *restrict input2, T *restrict output, int32_t cols, float epsilon)
+void rms_norm_general(const T *restrict input,
+                      const T *restrict input2,
+                      T *restrict output,
+                      int32_t cols,
+                      float epsilon)
 {
     event0();
     ::aie::vector<float, N> add_res = ::aie::zeros<float, N>();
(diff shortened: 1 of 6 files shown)


aie_kernels/aie2p/rms_norm.cc
```diff
@@ -33,19 +32,22 @@ void rms_norm_general(const T *restrict input, const T *restrict input2, T *rest
 
     float rms = sum_sq / cols + epsilon;
     float inv_rms = aie::invsqrt(rms);
-    ::aie::vector<T, N> inv_rms_v = ::aie::broadcast<T, N>(static_cast<T>(inv_rms));
+    // Normalize in f32 and round once. The previous code cast inv_rms to bf16 and did a
+    // bf16*bf16 multiply, a coherent per-norm scale error that accumulated over a deep
+    // residual stream and flipped near-tie argmaxes. Mirrors layer_norm.cc's accfloat path.
```
andrej:
```suggestion
    // Normalize in f32 and round once. Mirrors layer_norm.cc's accfloat path.
```

aie_kernels/aie2/rms_norm.cc
```diff
@@ -35,19 +34,22 @@ void rms_norm_general(const T *restrict input, const T *restrict input2, T *rest
 
     float rms = sum_sq / cols + epsilon;
     float inv_rms = invsqrt(rms);
-    ::aie::vector<T, N> inv_rms_v = ::aie::broadcast<T, N>(static_cast<T>(inv_rms));
+    // Normalize in f32 and round once. The previous code cast inv_rms to bf16 and did a
+    // bf16*bf16 multiply, a coherent per-norm scale error that accumulated over a deep
+    // residual stream and flipped near-tie argmaxes. Mirrors layer_norm.cc's accfloat path.
```
andrej:
```suggestion
    // Normalize in f32 and round once/ Mirrors layer_norm.cc's accfloat path.
```

aie_kernels/aie2/rms_norm.cc
```diff
@@ -67,13 +69,15 @@ void rms_norm_general(const T *restrict input, const T *restrict input2, T *rest
 }
 
 extern "C" {
-void rms_norm_bf16_vector(bfloat16 *input, bfloat16 *output, int32_t size)
+void rms_norm_bf16_vector(bfloat16 *input, bfloat16 *output, int32_t size, float epsilon)
 {
-    rms_norm_general<bfloat16, 16>(input, nullptr, output, size);
+    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even; do not inherit a floor rounding mode from a prior kernel
```
andrej:
```suggestion
    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even
```

aie_kernels/aie2/rms_norm.cc
```diff
@@ -67,13 +69,15 @@ void rms_norm_general(const T *restrict input, const T *restrict input2, T *rest
...
+    rms_norm_general<bfloat16, 16>(input, nullptr, output, size, epsilon);
 }
 
-void weighted_rms_norm(bfloat16 *a_in, bfloat16 *b_in, bfloat16 *c_out, int32_t size)
+void weighted_rms_norm(bfloat16 *a_in, bfloat16 *b_in, bfloat16 *c_out, int32_t size, float epsilon)
 {
-    rms_norm_general<bfloat16, 16>(a_in, b_in, c_out, size);
+    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even; do not inherit a floor rounding mode from a prior kernel
```
andrej:
```suggestion
    ::aie::set_rounding(aie::rounding_mode::conv_even);  // round-to-nearest-even
```

andrej (approved):
Thanks for the accuracy improvements! Cleaned up the comments a bit (sorry for being anal), if this passes ready to merge, feel free to remind me if I forget again with everything else going on.

atassis:
Fixed CI

atassis:
Oh, and yes, no issues, they are all reasonable:)
I'll admit why I don't clean up everything by myself- sometimes even after proof reading the code and comments I tend to internally approve smth even after AI review and else. So it is easier for me to suggest smth, get a review and then fix that.
I do that because I find it easier for someone to have a fast look and direct me, instead of talking more time to try and deduce myselves. 
For example, after talking to Joseph I have found another idea, which led to me through few more bugs, which you have seen patched already. And that idea is not fully researched yet

PR #142, opened 2026-07-25
operators: emit trace configuration when trace_size is set

atassis:
Most designs in `iron/operators` accept a `trace_size` argument but never call
`Runtime.enable_trace`, so they run fine and return a `trace.txt` of all zeros. That reads as
"this op can't be traced" rather than "the wiring is missing", which is what sent me looking.

Affected: `axpy`, `binary_elementwise`, `channeled_unary` (so gelu/silu/relu/sigmoid/tanh),
`gemm`, `leaky_relu`, `mem_copy`, `mha`, `rope`, `softmax`.

Rather than repeat the same event setup per design, the wiring goes in
`iron/operators/_trace.py`. `maybe_enable_trace()` uses an explicit `trace_size` when one is
passed and otherwise falls back to `IRON_TRACE_SIZE`, so the designs whose `op.py` doesn't plumb
`trace_size` through can still be traced; `IRON_TRACE_NTILES` caps how many workers are
instrumented (0 traces none). It's a no-op when neither is set, so default behaviour is unchanged.

## Three things I'd rather flag than have you find

**The import is absolute, not relative, and that matters.** `DesignGenerator` loads these files by
path via `spec_from_file_location` + `exec_module` with no package context
(`iron/common/compilation/base.py`), so a relative import raises "attempted relative import with no
known parent package" at generation time. It also passes `py_compile`, so it won't show up in a
lint pass. Worth knowing before anyone tidies it. This is also the first dependency from a
`design.py` onto the surrounding `iron` package -- if you'd rather designs stayed standalone, say
so and I'll inline the helper instead.

**`dequant` is deliberately untouched.** It already wires trace, but in a simpler form (no
`workers`, no `coretile_events`), so folding it onto the helper would change its behaviour rather
than just deduplicate. Happy to do it as a follow-up if you want one path.

**No automated coverage for the fix itself.** The verification below is a manual check; nothing in
CI would catch this regressing again. If you want a test asserting a design emits `aie.trace` when
`trace_size` is set, tell me the shape you'd prefer and I'll add it.

## Testing

Generated `softmax` through that same package-less loader: **0 `aie.trace` ops before this change
with tracing requested, 18 after**, and still 0 when tracing is off. The 18 are a full
`aie.trace @trace_core_1` block (8 core events, 3 port configs, start/stop broadcasts,
`host_config {buffer_size = 8192}`).

On device (Strix, NPU2): 935 tests pass across `gelu`, `softmax`, `axpy`, `mem_copy`, `rope`.
`black --check` and `reuse lint` both clean.

### DIFF
--- a/iron/operators/_trace.py
+++ b/iron/operators/_trace.py
@@ -3,30 +3,10 @@
 
 """Shared per-op NPU hardware-trace wiring for the iron/operators designs.
 
-Why this exists
----------------
-Every design in this package takes a ``trace_size`` argument, but a design only actually
-emits trace configuration if it calls ``Runtime.enable_trace``. Designs that accept
-``trace_size`` and never make that call compile and run fine and then hand back a
-``trace.txt`` full of zeros, which reads as "this op is untraceable" rather than
-"nobody wired it up".
-
-Before this helper the wiring had drifted three ways: ``channeled_unary_design`` honored
-the ``trace_size`` parameter *or* the ``IRON_TRACE_SIZE`` env var and passed a full
-core-tile event set; ``gemm/design`` honored *only* the env var, so passing
-``trace_size=`` did nothing; ``dequant/design`` honored only the parameter and passed no
-events at all. Nine other designs had no wiring whatsoever. Keeping one implementation
-here is what stops that from happening again.
-
-Semantics (the ``channeled_unary`` behaviour, which was the most complete):
+Semantics:
   * explicit ``trace_size`` wins; otherwise fall back to ``IRON_TRACE_SIZE``
   * ``IRON_TRACE_NTILES`` (default 1) caps how many workers get traced; 0 traces none
   * no-op when neither is set, so production paths are unaffected
-
-Note this is deliberately NOT the upstream mechanism being reimplemented:
-``Runtime.enable_trace`` is the same API upstream ships and upstream's own designs
-(``aie.iron.algorithms.reduce`` / ``transform``) already call it. This package is
-fork-carried, so the gap was ours alone.
 """
 
 import os
--- a/iron/operators/_trace.py
+++ b/iron/operators/_trace.py
@@ -1,0 +1,94 @@
+# SPDX-FileCopyrightText: Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared per-op NPU hardware-trace wiring for the iron/operators designs.
+
+Why this exists
+---------------
+Every design in this package takes a ``trace_size`` argument, but a design only actually
+emits trace configuration if it calls ``Runtime.enable_trace``. Designs that accept
+``trace_size`` and never make that call compile and run fine and then hand back a
+``trace.txt`` full of zeros, which reads as "this op is untraceable" rather than
+"nobody wired it up".
+
+Before this helper the wiring had drifted three ways: ``channeled_unary_design`` honored
+the ``trace_size`` parameter *or* the ``IRON_TRACE_SIZE`` env var and passed a full
+core-tile event set; ``gemm/design`` honored *only* the env var, so passing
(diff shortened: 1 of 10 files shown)


iron/operators/_trace.py
```diff
@@ -0,0 +1,94 @@
...
+Semantics (the ``channeled_unary`` behaviour, which was the most complete):
+  * explicit ``trace_size`` wins; otherwise fall back to ``IRON_TRACE_SIZE``
+  * ``IRON_TRACE_NTILES`` (default 1) caps how many workers get traced; 0 traces none
+  * no-op when neither is set, so production paths are unaffected
+
+Note this is deliberately NOT the upstream mechanism being reimplemented:
+``Runtime.enable_trace`` is the same API upstream ships and upstream's own designs
+(``aie.iron.algorithms.reduce`` / ``transform``) already call it. This package is
+fork-carried, so the gap was ours alone.
+"""
```
andrej, line 10:
```suggestion
"""Shared per-op NPU hardware-trace wiring for the iron/operators designs.

Semantics:
  * explicit ``trace_size`` wins; otherwise fall back to ``IRON_TRACE_SIZE``
  * ``IRON_TRACE_NTILES`` (default 1) caps how many workers get traced; 0 traces none
  * no-op when neither is set, so production paths are unaffected
"""
```

Here is something I have in my `AGENTS.md` -- not sure if it actually helps much but thought I'd share:
```
# Commenting style (applies to ALL code I write, every session)

Comments are expensive: they draw the reader's limited attention. Spend that
attention only on what the code cannot say for itself. Default to FEWER comments.

## A comment must earn its place
Write one ONLY to capture something not evident from the code itself:
- a *why* / rationale, a constraint, an invariant, a gotcha, a non-obvious
  consequence, or genuinely complex/surprising behavior.
Do NOT write a comment that restates what the adjacent code plainly does.

## Never narrate the change or the request
Code describes what IS, never its own history. The reader has no idea (and does
not care) what the code used to be, or that an edit happened. Do NOT:
- Turn my prompt / the change request into a comment.
- Document that code was previously wrong, broken, non-idiomatic, or slow.
- Explain that something "was changed", "now does X instead of Y", or "was
  refactored to follow <abstraction>". No changelog/diff narration in code —
  that belongs in the commit message, not the source.
- Reference a "baseline", "legacy", "old", "original", "previous", or "existing"
  path/behavior — or say a change "preserves"/"keeps"/"still" does something.
  These all frame the code against a prior version the reader can't see. Describe
  the code as it stands, with no before/after framing. (If two code paths
  genuinely coexist, name them for what they ARE, not which came first.)
- Justify following the codebase's normal conventions (that's just the default).

## Specific anti-patterns to avoid
- Paraphrasing the next line(s) of code.
- Enumerating where code is NOT (e.g. "parsing lives here, not in the driver").
- Long narrative blocks that duplicate the logic directly below them.
- Restating a well-named function/variable's obvious purpose.

## Default
When in doubt, delete the comment and let the code speak. Match the file's
existing comment density, but err toward less. Keep only high-signal comments.

```

iron/operators/_trace.py
```diff
@@ -0,0 +1,94 @@
...
+__all__ = ["maybe_enable_trace", "resolve_trace_size"]
+
+
+def resolve_trace_size(trace_size=None):
+    """Effective trace size: explicit argument first, then ``IRON_TRACE_SIZE``, else 0."""
+    if trace_size and trace_size > 0:
+        return int(trace_size)
+    # Deliberately unguarded: a malformed IRON_TRACE_SIZE should raise rather than
+    # silently disable tracing, which would reproduce the all-zeros trace.txt
+    # confusion this module exists to remove.
```
andrej:
```suggestion
    # silently disable tracing.
```

andrej (changes requested):
Thank you for cleaning up our messes. :)

iron/operators/_trace.py
atassis: Thanks, this is gonna be handy! It might help me with my notes, too

andrej (approved):

PR #150, opened 2026-08-14
Update mlir-aie to latest nightly (1.4.2.dev) and reduce IRON/mlir-aie duplication

hunhoffe:
- Bumps `mlir_aie` to the latest nightly (`1.4.2.dev...`, on top of #145's v1.4.0 migration) and `llvm-aie` to `22.0.0.2026081401` (LLVM 21→22).
- Fixes the resulting breaking changes: aiecc flipped its Peano/Chess default (`--no-xchesscc`/`--no-xbridge` no longer exist), and `XRTTensor`'s residency tracking was reworked into a shared storage/coherence-map model.
- Removes `--dynamic-objFifos` everywhere it was passed — it's the default upstream now.
- While fixing the coherence-tracking break, found that upstream added a first-class `subview()` API that fully supersedes our hand-rolled `XRTSubBuffer` class (it even resolves a FIXME we had: `XRTSubBuffer.to()` used to resync the *entire* parent buffer, clobbering sibling sub-views). Migrated all six call sites and deleted the class.
- One more small dedup: `get_kernel_dir()`'s device→arch-string mapping now delegates to upstream's `resolve_target_arch()` instead of reimplementing it.
- Also fixes a pre-existing bug found along the way: `iron/operators/__init__.py` unconditionally imported the stream-dse-backed operator, which broke its documented self-skip behavior when `onnx`/`stream-dse` aren't installed.

## PR Merge Checklist

1. [ ] The PR is rebased on the latest `devel` commit and pointing to `devel`.
2. [ ] Your PR has been reviewed and approved.
3. [ ] All checks are passing.

PR #154, opened 2026-08-26
Wire the AIE trace buffer through the fused full ELF flow

asyms:
stream-dse can now emit AIE tracing, but the trace buffer never reaches the device through the fused full ELF flow: lowering appends it to the operator's runtime sequence, and the address gets patched against the dispatched kernel instead, which is the wrapper. This wires the buffer through so a traced design produces real data, and leaves untraced builds exactly as they were.

- Give the fused runtime sequence a trace buffer per operator, at the index that operator uses. That index is what lowering patches against, so the wrapper has to match it or the trace DMA writes nowhere and the buffer comes back empty.
- Allocate one buffer per operator, bind them, and sync them back with the outputs.
- Let the SwiGLU-prefill operator use it: it passes `trace_size` into the sequence, and `trace_size` and `trace_max_tiles` into stream-dse's codegen, so the design is generated with tracing rather than having it added afterwards.
- Turn tracing on with `IRON_TRACE_SIZE`, and pick how many tiles with `IRON_TRACE_NTILES`. Off by default, since tracing changes the calling convention and its packet flows need routing a full design may not have spare.
- Key both the generated design and the sequence's ELF on whether tracing is on, so traced and untraced builds never share an entry. Without the second one, turning the flag on reuses an untraced ELF and binding the trace buffer fails against a kernel that has no argument for it.
- Require stream-dse 1.13.14, the first release with the tracing codegen and the tile count argument. This is a floor for every build, not just traced ones.
- Add unit tests for the slot layout.

This reopens #151, which targeted `update-mlir-aie-latest-nightly` and was closed when that branch merged as #150. It is based on `devel` now, and carries one extra commit for the ELF cache key above.

Checked on Strix in a clean venv built exactly as `.github/actions/prereqs` does, against stream-dse 1.14.0 (what the `>=1.13.14` floor resolves to today) and mlir-aie 1.4.2.dev16: `pytest -m "not extensive" iron/operators/ iron/tests/` gives 990 passed, a traced k=1 build passes on hardware, and turning tracing on and off again in one build directory works both ways. `black --check` and `reuse lint` are clean.


iron/common/tracing_utils.py
```diff
@@ -0,0 +1,367 @@
...
+    mlir-aie's trace parser recovers which tiles and events were traced by scanning
+    for ``aiex.npu.write32`` ops and pattern-matching the trace-unit config
+    addresses. It does not understand the declarative ``aie.trace`` ops a design is
+    written with, so the module handed to aiecc is useless to it: those ops only
+    become register writes inside aiecc, in ``aie-insert-trace-flows``. What we want
+    is one of aiecc's own intermediates, which it leaves in the work dir beside the
+    source (``<source>.mlir.d/``).
+
+    Picks the file with the most ``write32`` ops rather than hardcoding a stage name,
+    since those names are aiecc's business and have changed before.
```
andrej, line 76: That seems like a hack ...

If the trace parser util needs a certain form of IR, we should explicitly request it to be generated using `aiecc --get-<whatever it needs>`. If it is not exposed as `--get-<whatever it needs>`, we should enable that in `aiecc`. Relying on intermediate outputs in `<source>.mlir.d` seems hacky; the compiler may choose to not generate them and do things all in-memory instead.

iron/common/compilation/sequence.py
```diff
@@ -34,6 +36,27 @@
...
+    if not trace_size:
+        return [0, 1, 2], {}, 3
+    trace_slots = dict(arg_counts)
+    counts = list(trace_slots.values())
+    shared = sorted({n for n in counts if counts.count(n) > 1})
+    if shared:
+        raise NotImplementedError(
+            "operators taking the same number of arguments would share one trace "
+            f"buffer (slots {shared}); trace them in separate dispatches"
+        )
```
andrej, line 54: Am I understanding correctly that this rejects anything where multiple operators in the same sequence take the same number of arguments?

andrej (changes requested):
Thanks for submitting this. I see two things that worry me:

- The error about "operators taking the same number of arguments".
- The "pick the IR file with the most write32s" logic.

I think the fundamental problem is that tracing currently does not play nice with `aiex.configure`. What I read from this PR is that tracing always patches a buffer at a fixed index, but we'd want separate indices for each sub-operator to have its own trace buffer. To me, this PR looks like mostly a workaround around that. 

I hate rejecting this work, but could we instead implement a proper fix upstream? For example, make trace aware of `aiex.configure` and allocate separate trace buffers / different offsets into the same trace buffer for each called runtime sequence. I think at the end of the day that will be fewer lines of code, and it will benefit everyone upstream.

As a solution in the meantime, have you tried the "separate xclbin" dispatch mode for tracing?

andrej:
I've sent off an agent to look into a proper upstream fix to help with this, will report back.

andrej:
Upstream WIP https://github.com/Xilinx/mlir-aie/pull/3652

asyms:
@andrej Thanks for the upstream PR, that's indeed the way to go I think. Those things you mentioned were indeed added as hacks around a cleaner upstream change. I can work more on this tomorrow, would you like me to work directly in that mlir-aie branch in your fork or what do you think is the most appropriate way forward?

andrej:
@asyms I haven't had a chance to review the WIP PR in-depth yet but skimming it the overall shape seems correct. If you want to push directly to my branch on my fork, that'd be great. I think timezones will sequentialize us so we won't step on each other's toes too much. I'm hoping to get this merged this week.

andrej:
Sorry, bad communication on my part, I already have #178 open that picks up the changes from upstream.

PR #157, opened 2026-08-26
test_utils: make exact equality expressible in verify_buffer

atassis:
`verify_buffer` thresholds on `diff >= np.maximum(abs_tol, rel_tol * norm)`. Set both
tolerances to 0 and the threshold is 0, so `diff >= 0` holds for every element of a
bit-identical buffer: two copies of the same tensor score 100% errors. Exact equality
is not a strict gate today, it is an unusable one, and it cannot tell a perfect result
from a broken one. `nearly_equal`, directly above it in the same file, has the explicit
`if a == b: return True` escape that the vectorised path never grew.

That matters for the operators that do no arithmetic — transpose, mem_copy,
strided_copy, repeat — which currently have no way to ask for the gate they want.

The comparison is now strict, which is also what `np.isclose` means by its tolerances.
The new mask is a subset of the old one, so nothing that passes today can start failing;
only elements exactly at the threshold change verdict, and at the default 1e-6 / 0.04
that is measure-zero on float data.

## Added
- `iron/tests/infrastructure/comparison.py`: an identical buffer passes at 0/0 in f32 and
  bf16, one wrong element is reported alone at both 0/0 and the defaults, and 0/0 still
  rejects a one-ulp error that the default absorbs. 50 passed here, 40 of the 50 fail
  against the unfixed helper.

## Changed
- `iron/common/test_utils.py`: `>=` to `>` in the threshold comparison.
- `conftest.py`: the CSV reporting hook required a `[...]` suffix on every nodeid and
  raised otherwise, so a test with no parameters aborted the session. On devel today,
  `pytest iron/tests/stream/names.py --iterations 1` dies with an INTERNALERROR on
  `test_module_parameters_cover_the_golden_weights`. Included here because the new file
  adds another unparametrized test; the bug predates it.

## Removed
-

## Scope
CI runs `iron/operators` and `iron/applications` and never passes `--iterations`, so
neither the conftest path nor the new test file is reachable from CI as configured.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -143,10 +143,16 @@ def pytest_runtest_makereport(item, call):
             nodeid_components = re.match(
                 r"^(.+?::[^\[]+)\[(iter\d+-)?(.+?)\]$", item.nodeid
             )
-            if not nodeid_components:
-                raise RuntimeError(f"Unexpected test nodeid format: {item.nodeid}")
-            test_path = nodeid_components.group(1)
-            test_name = nodeid_components.group(3)
+            if nodeid_components:
+                test_path = nodeid_components.group(1)
+                test_name = nodeid_components.group(3)
+            else:
+                # A test with no parameters carries no [...] suffix at all. The
+                # iteration parametrize that would otherwise supply one is only
+                # added when --iterations > 1, so this is reachable at
+                # --iterations 1 and used to abort the whole session.
+                test_path = item.nodeid
+                test_name = item.nodeid.rsplit("::", 1)[-1]
 
             passed = report.outcome == "passed"
             captured = report.capstdout
--- a/conftest.py
+++ b/conftest.py
@@ -147,10 +147,8 @@ def pytest_runtest_makereport(item, call):
                 test_path = nodeid_components.group(1)
                 test_name = nodeid_components.group(3)
             else:
-                # A test with no parameters carries no [...] suffix at all. The
-                # iteration parametrize that would otherwise supply one is only
-                # added when --iterations > 1, so this is reachable at
-                # --iterations 1 and used to abort the whole session.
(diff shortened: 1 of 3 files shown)


iron/common/test_utils.py
```diff
@@ -99,7 +99,11 @@ def _to_numpy(x):
         + np.abs(expected_np[:compare_len].astype(float)),
         np.finfo(np.float32).max,
     )
-    mask = diff >= np.maximum(abs_tol, rel_tol * norm)
+    # Fail on diff STRICTLY past the threshold. With >=, rel_tol=abs_tol=0 makes the
+    # threshold 0 and every element fails, so a bit-identical buffer scores 100% errors
+    # and exact equality -- the honest gate for an operator that does no arithmetic --
+    # cannot be expressed at all.
```
andrej:
```suggestion
    # Use `>`, not `>=`, here, so that a user can pass rel_tol=abs_tol=0 
    # check exact equality.
```

conftest.py
```diff
@@ -143,10 +143,16 @@ def pytest_runtest_makereport(item, call):
...
-            test_path = nodeid_components.group(1)
-            test_name = nodeid_components.group(3)
+            if nodeid_components:
+                test_path = nodeid_components.group(1)
+                test_name = nodeid_components.group(3)
+            else:
+                # A test with no parameters carries no [...] suffix at all. The
+                # iteration parametrize that would otherwise supply one is only
+                # added when --iterations > 1, so this is reachable at
+                # --iterations 1 and used to abort the whole session.
```
andrej:
```suggestion
                # A test with no parameters carries no [...] suffix, so won't
                # match the regex above.
```

andrej (approved):

PR #158, opened 2026-08-26
strided_copy: fix the num_aie_channels>1 device hang, and add test coverage

atassis:
Stacked on #157: the new tests gate at exact equality, which #157 is what makes
expressible. Its two commits are in this diff and drop out when it merges.

`transfer_size` defaults to the product of the whole `input_sizes`, but the channel
split happens afterwards, so at `num_aie_channels > 1` the ObjectFifo object is N times
the length of the shim BD feeding it. A BD shorter than its object starves the MemTile's
S2MM: it never completes an object, never releases the lock, the linked MM2S never
starts, and the drain's `dma_await_task` never returns. The host sees
`ERT_CMD_STATE_TIMEOUT`. Generation, the MLIR verifier and the xclbin build are all clean.

An integer multiple is fine — it just cycles the buffer. Stated as that ratio, it
predicted 8 of 8 arms on a sweep across channels 1/2/4 x transfer_size. `num_aie_channels=4`
lands on both sides — passes at `transfer_size=256`, hangs at 512 — which is what rules
out a multi-channel defect and leaves the ratio.

The operator also had no tests, and could not have had: `get_arg_spec` passed a bare int
where `AIERuntimeArgSpec.shape` is a tuple, and `XRTTensor` only treats a tuple as a shape.
Anything else goes through `np.asarray`, so the int becomes a 0-d array and the output
buffer is allocated for one element. It is the only one of the tree's 35 arg-spec call
sites that passes a scalar.

## Added
- `iron/operators/strided_copy/test.py` and `reference.py`. Arms cover contiguous, chunked,
  2 and 4 channels, and Llama's KV-cache write shape at several slots, plus a construction-time
  rejection test. Gated at `rel_tol=abs_tol=0` — the operator does no arithmetic.

## Changed
- `design.py`: size the object against the per-channel share, and assert the divisibility
  the hardware requires, so ratio < 1 is a construction-time rejection instead of a device
  hang. Also assert that a copy moves the same element count both ways, which nothing
  checked and which fails the same undiagnosable way on the output side.
- `op.py`: the arg spec now passes a shape tuple.

## Removed
-

## Evidence
On devel with `design.py` reverted, exactly the four multi-channel arms fail with
`ERT_CMD_STATE_TIMEOUT` and the single-channel and chunked arms pass. With the fix,
11/11 pass at exact equality. Arithmetic at `num_aie_channels=1` is unchanged, so
`llama_npu.py`'s KV-cache write — the only in-tree caller — is untouched.

## Known gap
`llama_npu.py` drives the output offset at runtime through a `ScratchpadParameter`
(`output_offset_parameter="cache_offset"`). The new arms bake the offset as a compile-time
constant instead, so that path is still uncovered.

### DIFF
--- a/iron/operators/strided_copy/test.py
+++ b/iron/operators/strided_copy/test.py
@@ -73,8 +73,7 @@ def get_params():
 )
 @pytest.mark.parametrize("kwargs", get_params())
 def test_strided_copy(kwargs, aie_context):
-    """StridedCopy moves data and computes nothing, so the gate is exact equality.
-    """
+    """StridedCopy moves data and computes nothing, so the gate is exact equality."""
     # transfer_size only sizes the ObjectFifo; it does not move the data anywhere else,
     # so the golden is computed without it.
     golden_kwargs = {k: v for k, v in kwargs.items() if k != "transfer_size"}
--- a/iron/operators/strided_copy/test.py
+++ b/iron/operators/strided_copy/test.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from iron.operators.strided_copy.op import StridedCopy
+from iron.operators.strided_copy.reference import generate_golden_reference
+from iron.common.test_utils import run_test
+
+# Llama's KV-cache write, shrunk: the cache is (n_kv_groups, seq, head_dim) and one
+# token's keys land in slot t of every group. SEQ is 128 rather than the real 2048 to
+# keep the output buffer at 128 KB; the full-size arm is extensive.
+N_KV, HEAD_DIM, SEQ = 8, 64, 128
+
+
+def _kv_slot(seq, slot, num_aie_channels=1):
+    """StridedCopy kwargs writing one (N_KV, HEAD_DIM) token into cache slot `slot`."""
+    return dict(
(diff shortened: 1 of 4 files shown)


iron/operators/strided_copy/test.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+
+
+def get_params():
+    return [
+        pytest.param(_flat(1024), id="contiguous"),
+        # num_aie_channels splits the highest dimension across that many shim channels.
+        # Sweep it at the default transfer_size and at a divisor of the per-channel share:
+        # the object has to divide the per-channel BD, and the default used to be sized
+        # against the whole tensor, which hung every channel count above 1.
```
andrej:
```suggestion
```

iron/operators/strided_copy/test.py
```diff
@@ -0,0 +1,119 @@
...
+    Bandwidth=r"Effective Bandwidth: (?P<value>[\d\.e\+-]+) GB/s",
+)
+@pytest.mark.parametrize("kwargs", get_params())
+def test_strided_copy(kwargs, aie_context):
+    """StridedCopy moves data and computes nothing, so the gate is exact equality.
+
+    The kv_slot arms write one token into a cache slot and leave the rest of the
+    buffer alone, which is the shape that matters: a tolerance gate would accept a
+    copy that landed in the wrong slot, and a wrong slot is what a mis-scaled offset
+    produces.
```
andrej:
```suggestion
```

andrej (approved):
Good fix, and thanks for adding tests to an operator that previously had none.

PR #159, opened 2026-08-26
transpose, mem_copy: gate the pure-movement operators exactly

atassis:
Stacked on #157, which is what makes `rel_tol=abs_tol=0` mean exact rather than "fail
everything". Its two commits are in this diff and drop out when it merges.

Neither operator computes anything — transpose is a permutation, mem_copy is a copy — so
the tolerances they run under (0.04 and 0.01) also accept some class of wrong answer.
For a permutation that class includes landing in the wrong place, which is exactly what
a mis-scaled offset or a mis-split dimension produces.

## Added
-

## Changed
- `iron/operators/transpose/test.py`, `iron/operators/mem_copy/test.py`: gate at
  `rel_tol=abs_tol=0`.

## Removed
-

## Evidence
96/96 pass on npu2, every arm including the extensive ones: 32 transpose, 64 mem_copy.

### DIFF
--- a/iron/operators/mem_copy/test.py
+++ b/iron/operators/mem_copy/test.py
@@ -78,7 +78,12 @@ def test_mem_copy(
     output_buffers = {"output": golden_ref["output"]}
 
     errors, latency_us, bandwidth_gbps = run_test(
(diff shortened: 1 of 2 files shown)


andrej (approved):
Good! Thanks!

PR #160, opened 2026-08-26
mem_copy: the passThrough kernel is compiled for the wrong element type

atassis:
`mem_copy`'s non-bypass path types its line buffers bf16 (`design.py:179`) and links
`aie_kernels/generic/passThrough.cc`, which picks its element type from a `BIT_WIDTH`
macro the caller must supply. `get_kernel_artifacts` passes no `extra_flags`, so
`BIT_WIDTH` is undefined, the preprocessor evaluates it as 0, and the `#else // 32`
branch compiles. `mha/op.py:105` is the same kernel done correctly, with
`extra_flags=["-DBIT_WIDTH=16"]`; those two are its only consumers.

The template copies one `v64uint8` -- 64 bytes -- per `N` elements, so `N` is what ties
the trip count to the element width: 32 for int16, 16 for int32. Compiled as
`passThrough_aie<int32_t, 16>` but handed a bf16 element count, it runs `width/16`
iterations of 64 bytes = `width * 4` bytes against a buffer holding `width * 2`.

Read off the built artifact rather than inferred:

    $ llvm-objdump -t build/mem_copy.o | grep passThrough_aie
    _Z15passThrough_aieIiLi16EEvPT_S1_ii          # passThrough_aie<int, 16>

and in the generated MLIR the objectFifo buffers are `memref<64xbf16>` -- 128 bytes --
while the call passes `lineWidth = 64`, so each one has 4 x 64 = 256 bytes written
through it. 128 bytes past the end, into whatever the allocator placed next.

## The second half, which the first was hiding

Setting the flag makes one arm hang, deterministically, 3 of 3:
`input_length=1024, num_cores=16` gives a 64-element tile, which at the correct N=32 is
**two** loop iterations -- against the `AIE_LOOP_MIN_ITERATION_COUNT(6)` the loop
declares. The int32 miscompile was running four, still under six but evidently enough to
survive, so the wrong element type was masking a false promise underneath it.

The kernel cannot make that promise: the trip count is `(height * width) / N` with all
three supplied by the caller. Removed rather than lowered.

## Added
-

## Changed
- `iron/operators/mem_copy/op.py`: pass `-DBIT_WIDTH=16`, matching the bf16 line type.
- `aie_kernels/generic/passThrough.cc`: drop the unverifiable minimum-trip-count promise.

## Removed
-

## Evidence
67/67 pass -- 64 mem_copy arms including the extensive ones, plus mha, the kernel's other
consumer. The object now carries `passThrough_aie<short, 32>` and each call moves 128
bytes.

64/64 mem_copy arms also passed *before* this change, which is the point rather than a
reassurance: the overrun is past the region the tests compare, so nothing in the suite
could see it.

Dropping the pipelining hint costs nothing measurable at the largest shape
(8192 elements, one core, 256 iterations): median 181.6 us against 177.7 us over 12 reps
each, inside a 140-245 / 123-491 spread. Per-dispatch overhead dominates that
measurement, so it bounds the cost rather than showing it is zero.

### DIFF
--- a/aie_kernels/generic/passThrough.cc
+++ b/aie_kernels/generic/passThrough.cc
@@ -10,7 +10,12 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-template <typename T, int N>
+// Pipelined selects the loop's minimum-trip-count promise. The kernel cannot prove a
+// trip count from height/width/N alone -- that has to come from the caller, since
+// asserting >=6 for a call that runs fewer iterations hangs on device (mem_copy's
+// 64-element bf16 tile runs two at N=32). Callers that can guarantee >=6 should use
+// the Pipelined=true entry points below for the software-pipelined loop.
+template <typename T, int N, bool Pipelined>
 __attribute__((noinline)) void
 passThrough_aie(T *restrict in, T *restrict out, const int32_t height, const int32_t width)
 {
@@ -19,13 +24,19 @@ passThrough_aie(T *restrict in, T *restrict out, const int32_t height, const int
     v64uint8 *restrict outPtr = (v64uint8 *)out;
     v64uint8 *restrict inPtr = (v64uint8 *)in;
 
-    AIE_PREPARE_FOR_PIPELINING
-    // No minimum trip count: the caller supplies all of height, width and N. The 6 that
-    // was asserted here is false for mem_copy's 64-element bf16 tile, which runs two
-    // iterations at N=32 and hangs on device.
-    for (int j = 0; j < (height * width); j += N) // Nx samples per loop
-    {
-        *outPtr++ = *inPtr++;
+    if constexpr (Pipelined) {
+        AIE_PREPARE_FOR_PIPELINING
+        AIE_LOOP_MIN_ITERATION_COUNT(6)
+        for (int j = 0; j < (height * width); j += N) // Nx samples per loop
+        {
+            *outPtr++ = *inPtr++;
+        }
+    } else {
+        AIE_PREPARE_FOR_PIPELINING
+        for (int j = 0; j < (height * width); j += N) // Nx samples per loop
+        {
+            *outPtr++ = *inPtr++;
+        }
     }
 
     event1();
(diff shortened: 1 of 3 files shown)


aie_kernels/generic/passThrough.cc
```diff
@@ -20,7 +20,9 @@ passThrough_aie(T *restrict in, T *restrict out, const int32_t height, const int
     v64uint8 *restrict inPtr = (v64uint8 *)in;
 
     AIE_PREPARE_FOR_PIPELINING
-    AIE_LOOP_MIN_ITERATION_COUNT(6)
+    // No minimum trip count: the caller supplies all of height, width and N. The 6 that
+    // was asserted here is false for mem_copy's 64-element bf16 tile, which runs two
+    // iterations at N=32 and hangs on device.
     for (int j = 0; j < (height * width); j += N) // Nx samples per loop
     {
```
andrej, line 25: This likely has a performance implication. Have you compared the generated assembly with and without? Could you please paste below?

If they don't change with/without this macro, fine to delete. If not, I'd like a specialization for >6 trip count and one without, so we can get max performance for large buffers. We use this for performance measurements of bandwidth, so it is important.

Also note

andrej (commented):
Good change in principle, but need to know about performance implications.

Also, would it be much extra work to allow other data types and pass the right bit width depending on input dtype in op.py?

And, one remark: We might end up deciding to hold PRs that modify kernels here until https://github.com/Xilinx/mlir-aie/pull/3599 is merged in MLIR-AIE, then ask to migrate the changes there rather than here. Just as a forward-looking warning.

atassis:
Assembly, both arches, all three widths, pinned Peano (llvm-aie 299581a415e0). Same source, only
the hint removed:

| arch | steady-state loop | zero-trip guard | function total |
|---|---|---|---|
| aie2p, hint | **2 bundles** | no | 25 |
| aie2p, no hint | **8 bundles** | yes | 31 |
| aie2, hint | **2 bundles** | no | 26 |
| aie2, no hint | **9 bundles** | yes | 31 |

Identical at BIT_WIDTH 8, 16 and 32. The inner loop is 4-4.5x longer without it.

With the hint the loop is software-pipelined -- `add.nc lc, r0, #-3`, loads hoisted into a
prologue, stores drained in the return delay slots -- and there is no trip-count guard. Without
it the compiler emits `mova r1,#0; ge r1,r1,r0; jnz r1` plus five delay-slot nops, drops the
pipelining (`add.nc lc, r0, #0`), and the body becomes one un-overlapped load and store.

So you were right and I was wrong to drop it. Taking your second option: a specialization that
keeps the hint for callers that can promise >6 trip counts, and an unhinted path for the ones
that cannot. `mem_copy`'s 64-element bf16 tile runs two iterations at N=32, which is where the
device hang came from.

One number in the other direction, so it is on the record: at the largest shape (8192 elements,
one core, 256 iterations) the wall-clock difference was 181.6 vs 177.7 us median over 12 reps
each, inside a 140-245 / 123-491 spread. Per-dispatch overhead dominates there, so that
measurement bounds nothing -- the bundle counts are the real signal.

Also, now that #3599 has merged: should this move to mlir-aie rather than land here?

andrej:
> So you were right and I was wrong to drop it. Taking your second option: a specialization that
keeps the hint for callers that can promise >6 trip counts, and an unhinted path for the ones
that cannot. mem_copy's 64-element bf16 tile runs two iterations at N=32, which is where the
device hang came from.

Great! That is the correct fix to not regress on performance.

> Also, now that #3599 has merged: should this move to mlir-aie rather than land here?

Yes, let's migrate this over to mlir-aie. @efurst is working on a PR that will remove the kernels from this repository, so adding changes here would get lost. Apologies for the extra work. I'll close this for now, please comment with the matching PR in MLIR-AIE once you have it opened there.

PR #161, opened 2026-08-26
repeat: split cols against every constraint the BD imposes, and add test coverage

atassis:
Stacked on #157: the new tests gate at exact equality, which #157 is what makes
expressible. Its two commits are in this diff and drop out when it merges.

`repeat` has no tests, and could not have had: `rows=8, cols=512, repeat=4` does not
build.

Both taps split `cols` as

    sizes=[repeat, rows, cols // cols_split, cols_split]   strides=[..., cols_split, 1]

so the innermost dim counts chunks and the contiguous run sits one dim out. `cols_split`
is 1 whenever no split is needed, which makes the innermost dim one element. At bf16
that is two bytes, not a whole 32-bit word, and `verifyStridesWraps` rejects it:

    'aie.dma_bd' op Transfer sizes must be multiples of 4 bytes. 1 elements at 2 bytes
    each equal 2 bytes, which is not divisible by 4.
    aie.dma_bd(%arg0 : memref<8x512xbf16> ... sizes = [4, 8, 512, 1] strides = [0, 512, 1, 1])

The 512-element stride-1 run the transfer is actually made of is right there in dim 2.

The reason this survived: `llama_npu.py` passes `cols = prompt_len * head_dim`, far above
1023, so it lands at `cols_split = 256` and its innermost dim is a healthy 256 elements.
The only caller in the tree never trips it; every small-`cols` shape does, which is what a
test would use.

## Added
- `iron/operators/repeat/test.py`. Five device arms at `rel_tol=abs_tol=0` (repeat moves
  data and computes nothing) spanning `cols_split` 1, 2, 4 and 256 -- the last being the
  shape `llama_npu.py` actually dispatches -- plus three arms pinning the shapes that have
  no legal split. 8/8.

## Changed
- `iron/operators/repeat/design.py`: swap the two inner dims so the innermost carries
  `cols // cols_split` at stride 1. The address sequence is byte-identical either way --
  the nested (outer, inner) walk of a linear range is the same walk regardless of which
  factor is outer -- so the change is only in which dim the hardware sees as innermost.

  The swap moves `cols_split` onto the next dim out, which carries the same 10-bit wrap
  field, so bounding the chunk length alone is not sufficient. The selection loop now
  searches for a divisor satisfying all three constraints at once: chunk <= 1023, chunk a
  whole number of 32-bit words, and count <= 1023. Where no divisor satisfies them -- an
  odd `cols`, a prime above 1023, or twice such a prime -- it raises rather than emitting a
  tap the BD verifier rejects less legibly. Every `cols` that worked before selects the
  same `cols_split`, including llama's 131072.

## Removed
-

## Evidence
8/8 on device, run without the `not extensive` filter so both extensive arms -- including
the llama shape -- actually executed. The emitted BD for the llama arm is
`sizes = [4, 8, 256, 512] strides = [0, 131072, 512, 1]`.

### DIFF
--- a/iron/operators/repeat/design.py
+++ b/iron/operators/repeat/design.py
@@ -12,20 +12,32 @@ from aie.iron import ObjectFifo, Program, Runtime, TaskGroup
 
 
 def repeat(dev, dtype, rows, cols, repeat, transfer_size=None):
+    elem_bytes = np.dtype(dtype).itemsize
     dtype = np.dtype[dtype]
 
-    # Try to work around hardware size limitations by breaking transfers into smaller chunks
-    cols_split = 1
-    if cols > 1023:
-        for divisor in range(2, cols + 1):
-            if cols % divisor == 0 and cols // divisor <= 1023:
-                cols_split = divisor
-                break
-        else:
-            raise ValueError(
-                f"Cannot split cols={cols} into chunks <= 1023; hardware limits cols to not exceed 1023"
-            )
-    assert cols_split <= 1023, "cols is too large, can't split into smaller transfers"
+    # Split cols into cols_split chunks of cols // cols_split. Both land on a BD dimension
+    # and each carries its own hardware constraint, so the search has to satisfy all three
+    # at once rather than the chunk length alone:
+    #   - the chunk length is the innermost dim: <= 1023 (10-bit wrap) AND a whole number
+    #     of 32-bit words, since the BD's innermost size is denominated in words
+    #   - the chunk count is the next dim out: <= 1023, the same wrap field
+    # An odd cols has only odd divisors, so no split of it is ever word-aligned at bf16;
+    # that is reported here rather than left to the BD verifier.
+    granule = max(1, 4 // elem_bytes)  # elements per 32-bit word
+    cols_split = None
+    for divisor in range(1, cols + 1):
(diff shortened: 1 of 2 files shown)


iron/operators/repeat/design.py
```diff
@@ -12,20 +12,32 @@
...
-                cols_split = divisor
-                break
-        else:
-            raise ValueError(
-                f"Cannot split cols={cols} into chunks <= 1023; hardware limits cols to not exceed 1023"
-            )
-    assert cols_split <= 1023, "cols is too large, can't split into smaller transfers"
+    # Split cols into cols_split chunks of cols // cols_split. Both land on a BD dimension
+    # and each carries its own hardware constraint, so the search has to satisfy all three
+    # at once rather than the chunk length alone:
```
andrej:
```suggestion
    # Split cols into cols_split chunks of cols // cols_split. This is required to 
    # satisfy hardware constraints on BD dimensions. We must choose a split that 
    # does not exceed the hardware register sizes:
```

andrej (approved):
Nice, this is a good improvement, and thanks for adding testing. I added these ops in a rush to enable llama, not really testing them (obviously) for other cases, so this is a great improvement and a clean solution with figuring out the right split.

PR #162, opened 2026-08-28
Copy aiecc work-dir inputs when Windows denies symlinks.

jtuyls:
aiecc looks up kernel objects in a per-artifact work dir. We currently symlink those files from the shared build dir. On Windows without Developer Mode, `Path.symlink_to` raises `OSError` and compile fails. Fall back to `shutil.copy2` so the same lookup still works.

## Added
- Copy fallback in `_link_build_outputs_into` when symlink creation is denied.

## Changed
- `_link_build_outputs_into` still prefers a symlink; it copies only on `OSError`.

## Removed
-

## PR Merge Checklist


andrej (approved):
Thanks for the clean, minimal fix!

PR #163, opened 2026-08-28
Re-export operator classes lazily.

jtuyls:
`iron.operators` bound every operator class in `__init__.py`. Importing one name, or even `import iron.operators.elementwise_add.op`, therefore imported the whole catalog: Python loads parent packages first, so the eager `__init__` ran before the operator you asked for.

Each `iron.operators.<name>.op` module does not need to import the others. PEP 562 `__getattr__` keeps the same public names and only loads the operator that was asked for. A broken or heavy import in one operator then cannot fail compile or use of another.

## Added
- PEP 562 `__getattr__` / `__dir__` re-exports for the existing operator class names.
- `iron/tests/infrastructure/lazy_imports.py`: `ElementwiseAdd` does not import MHA or SwiGLU.

## Changed
- `iron.operators` public names are unchanged; they are imported on first access.

## Removed
- Eager `from .mha.op import MHA` (and the other catalog imports) from `iron/operators/__init__.py`.

## PR Merge Checklist

### DIFF
--- a/iron/operators/__init__.py
+++ b/iron/operators/__init__.py
@@ -1,17 +1,42 @@
 # SPDX-FileCopyrightText: Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from .elementwise_add.op import ElementwiseAdd
-from .elementwise_mul.op import ElementwiseMul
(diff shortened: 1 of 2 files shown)


andrej (approved):
Great, makes sense, learned something today, thanks!

PR #164, opened 2026-08-28
Allow compiling MHA and OperatorSequence without torch.

jtuyls:
Compile and NPU dispatch do not use torch. MHA imported it at module load for padding, and `OperatorSequence` imported it for CPU reference/compare. Move torch off the compile path so those modules import with compiler-stack only. Tests and CPU goldens still use torch via `requirements.txt`.

## Added
- NumPy pad for MHA sequence padding.
- Lazy `_torch()` import on the sequence CPU reference/compare paths.

## Changed
- `pyproject.toml` package deps are numpy and ml_dtypes. Torch stays in `requirements.txt` for tests.

## Removed
- `import torch` from `iron/operators/mha/op.py` and from `iron.common.sequence` module load.
- Torch as a hard `pyproject.toml` dependency.

## PR Merge Checklist

### DIFF
--- a/iron/common/sequence.py
+++ b/iron/common/sequence.py
@@ -8,7 +8,6 @@ from pathlib import Path
 import numpy as np
 import ml_dtypes
 import pyxrt
-import torch
 from . import compilation as comp
 from .base import AIEOperatorBase, MLIROperator
 import aie.utils as aie_utils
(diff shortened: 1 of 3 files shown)


andrej (approved):
Nice to drop a heavy dependency like this for some users, though it'll still be needed for a lot of the other operators

PR #165, opened 2026-08-28
Allow compiling OperatorSequence without XRT.

jtuyls:
`OperatorSequence` imported `pyxrt` and `XRTTensor` at module load, so compiling SwiGLU (or any sequence) required XRT even though compile and the CPU dispatch path do not use it. Make those imports optional. XRT-native dispatch policies still need pyxrt and fail at construction if it is missing. `run_test` uses `aie.utils.DEFAULT_TENSOR_CLASS` so operator tests follow `IRON_RUNTIME` instead of hard-coding XRT.

## Added
- Optional `pyxrt` / `XRTTensor` import in `iron.common.sequence`.
- `_require_xrt()` on the full-ELF and chained-xclbin callables.

## Changed
- `run_test` allocates buffers with `DEFAULT_TENSOR_CLASS` and measures payload size from the numpy view.

## Removed
- Unconditional `import pyxrt` and `XRTTensor` from `sequence.py` and `test_utils.py`.

## PR Merge Checklist

### DIFF
--- a/iron/common/test_utils.py
+++ b/iron/common/test_utils.py
@@ -8,7 +8,6 @@ import torch
 import aie.utils as aie_utils
 from ml_dtypes import bfloat16
 from .base import AIEOperatorBase
-from aie.utils.hostruntime.xrtruntime.tensor import XRTTensor
 
 torch_dtype_map = {
     "bf16": torch.bfloat16,
@@ -126,6 +125,17 @@ def verify_buffer(
     return errors
 
 
+def _nbytes(buf) -> int:
+    """Bytes of tensor data moved, for the effective-bandwidth figure.
+
+    Reads the numpy view rather than the backend buffer handle: ``buffer_object()``
+    returns a ``pyxrt.bo`` under XRT but an opaque handle under HRX, so ``.size()`` is
+    not part of the Tensor interface. The view is also the more honest number -- it is
+    the payload, not the (page-rounded) allocation.
+    """
+    return buf.data.nbytes
+
+
 def run_test(
     operator: AIEOperatorBase,
     input_buffers: dict[str, torch.Tensor],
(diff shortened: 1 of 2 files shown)


iron/common/test_utils.py
```diff
@@ -167,34 +177,41 @@ def run_test(
 
     total_bytes = 0
 
+    # The device tensor type of whichever host runtime is selected (IRON_RUNTIME):
+    # XRTTensor under XRT, HRXTensor under HRX. Both implement the Tensor interface
+    # this function uses, and the operator dispatches through DefaultNPURuntime, which
+    # is the matching runtime -- so naming one backend's class here is what made the
+    # test suite XRT-only.
```
andrej:
```suggestion
    # is the matching runtime.
```

documents code history inline with code, confusing and irrelevant to readers not familiar with the history

andrej (approved):
This makes sense to me.

Two general HRX follow-up questions:

1. Are there plans to enable the full ELF path in HRX? Full ELF enables the `load_pdi` instruction using relocations to the PDI at runtime. Using `load_pdi` enabled significant speedups (2x+) for some of our applications, because it allows us to run multiple operators back-to-back without CPU interaction. It doesn't necessarily need to be full ELF either, although that would be easiest -- a runtime just needs to enable `load_pdi` by loading and relocating the memory address of the PDI to load the inside the `load_pdi` instruction.
2. There is a separate-xclbin "dispatch mode" for `OperatorSequence`. Could we enable using this today to use HRX with relatively minimal effort by using the HostRuntime abstraction? The "auto" dispatch mode could even automatically switch to "separate-xclbin" mode if XRT is missing but HRX is detected -- it would (for reasons in 1.) be slower but functional.

iron/common/test_utils.py
jtuyls: Indeed, thanks for catching that.

PR #167, opened 2026-08-29
gemv: add optional fused RMSNorm/LayerNorm prologue

atassis:
GEMV can now fold an RMSNorm or LayerNorm over its shared B vector, so a caller that
normalizes before a matvec stops paying a separate round trip for it.

## Added
- `prologue={"none","rms","ln"}` on the GEMV operator, applied to the shared B vector once
  per acquire, before it feeds every matvec call in the batch — so the cost is O(K) per
  batch, not O(M) as re-normalizing per output tile would be.
- `rms_norm_ref` / `layer_norm_ref` in `gemv/reference.py`, affine-free and computed in
  float32 to match the kernels' own reduction.
- `test_gemv_norm_prologue`, checking the fused result against an `A @ norm(B)` golden
  across both norms and three shapes.
- `layer_norm_f32` in `aie_kernels/aie2{,p}/layer_norm.cc`: an `extern "C"` entry point for
  the `layer_norm_bf16_f32_calculation` implementation that was already in the file but not
  exported.

## Changed
- The GEMV prologue calls `layer_norm_f32` rather than `layer_norm`. `layer_norm`
  accumulates the sum in bfloat16 (`::aie::vector<T, N> sum_acc`) while accumulating the
  sum of squares in f32, so the mean it subtracts loses accuracy as `cols` grows. That is
  visible here and not in RMSNorm, which needs no mean: against the f32 reference, `ln` at
  K=2048 missed on ~180 of 8192 outputs and at K=128 on one, while `rms` matched at the
  same shapes. `layer_norm` itself is untouched, so nothing else changes behavior.

## Removed
- Nothing.

## Known gap
The prologue reads and writes separate pointers, so it needs a second K-element L1 buffer
on top of the B fifo's. At K=8192 with `m_input=1` that buffer alone overflows one core's
data memory, and the failure surfaces as `Basic sequential allocation also failed` from
aiecc rather than as a precondition. The sweep is bounded to K=4096 accordingly; a real
up-front check wants an L1 size the device model exposes, which IRON does not carry today.

## PR Merge Checklist

### DIFF
--- a/iron/operators/gemv/design.py
+++ b/iron/operators/gemv/design.py
@@ -8,7 +8,7 @@ import aie.dialects.index as index
 from aie.dialects.aie import T
 from aie.helpers.dialects.scf import _for as range_
 from aie.helpers.taplib import TensorAccessPattern
-from aie.iron import Kernel, ObjectFifo, Program, Runtime, TaskGroup, Worker
+from aie.iron import Buffer, Kernel, ObjectFifo, Program, Runtime, TaskGroup, Worker
 
 """
 Matrix-vector design
@@ -37,6 +37,7 @@ def my_matvec(
     func_prefix="",
     verbose=False,
     epilogue="none",
+    prologue="none",
 ):
     if m_output is None:
         m_output = m_input
@@ -101,6 +102,26 @@ def my_matvec(
             [np.int32, L1_C_ty],
         )
 
+    # Optional norm applied to the shared B vector once per acquire, before it feeds every
+    # matvec call in this batch -- so the cost is O(K) per batch, not O(M) as re-normalizing
+    # per output tile would be. Both kernels take separate in/out pointers (restrict on
+    # both), so normalizing needs a second L1 buffer rather than an in-place call.
+    assert prologue in ("none", "rms", "ln")
+    norm_kernel = None
+    b_norm_bufs = [None] * cols
+    PROLOGUE_EPSILON = 1e-5  # matches layer_norm.cc's hardcoded epsilon
+    if prologue != "none":
+        norm_fn = "rms_norm_bf16_vector" if prologue == "rms" else "layer_norm"
+        norm_args = (
+            [L1_B_ty, L1_B_ty, np.int32, np.float32]
+            if prologue == "rms"
+            else [L1_B_ty, L1_B_ty, np.int32]
+        )
+        norm_kernel = Kernel(
+            f"{func_prefix}{norm_fn}", f"{func_prefix}{kernel_object}", norm_args
+        )
+        b_norm_bufs = [Buffer(type=L1_B_ty, name=f"B_norm_{i}") for i in range(cols)]
+
     A_L3L1_fifos = [
         ObjectFifo(L1_A_ty, name=f"A_L3L1_{i}", depth=2) for i in range(cols)
     ]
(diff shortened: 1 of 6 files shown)


iron/operators/gemv/design.py
```diff
@@ -101,6 +102,26 @@ def my_matvec(
             [np.int32, L1_C_ty],
         )
 
+    # Optional norm applied to the shared B vector once per acquire, before it feeds every
+    # matvec call in this batch -- so the cost is O(K) per batch, not O(M) as re-normalizing
+    # per output tile would be. Both kernels take separate in/out pointers (restrict on
+    # both), so normalizing needs a second L1 buffer rather than an in-place call.
```
andrej, line 108: Should we add an in-place activation kernel in the future? Seems wasteful ...

iron/operators/gemv/op.py
```diff
@@ -32,6 +32,10 @@ class GEMV(MLIROperator):
     # "none" (default) leaves the output unchanged; "gelu" applies GELU(tanh approx).
     # repr=False keeps operator/artifact names stable for the default path.
     epilogue: str = field(default="none", repr=False)
+    # Optional norm applied to the shared vector once before the matvec, fusing a
+    # RMSNorm/LayerNorm decode prologue into the same dispatch. Affine-free (gamma/beta
+    # fold into the weight matrix host-side, same convention the norm kernels already use).
```
andrej:
```suggestion
    # RMSNorm/LayerNorm decode prologue into the same dispatch. Affine-free (gamma/beta
    # fold into the weight matrix host-side).
```

andrej (approved):
A nicer "Kernel" abstraction that encodes (i) the function name (extern C), (ii) the argument specification expected (MLIR func.func signature), (iii) how to compile and link it in would be good at some point in the future. If we have to list every single activation function in every GEMV and GEMM kernel as a custom prologue/epilogue, it will get old quick.

But for now this is good. Thanks!

andrej:
One warning:

```
  /workspace/runner/_work/IRON/IRON/build/GEMV_M2048_K4096_col1_tsi1_tso2048_batch1_npu2_proln.mlir.d/aie.mlir:13:10: warning: cannot determine this core's stack requirement: no stack size information for 'layer_norm_f32' -- it is called (directly, or conservatively through a function-pointer reference) but its defining object was not measurable (missing .stack_sizes, an archive/bitcode input, or not part of this core's link_files); stack_size is not being validated for this core. Set stack_size_override on the affected kernel's external_func()/func.func declaration (Kernel(...)/ExternalFunction(...) in IRON) to enable it
```

And these errors:

```
  _ test_gemv_norm_prologue[iter0-M_2048-K_4096-num_aie_columns_1-tile_size_input_1-tile_size_output_2048-prologue_ln-norm_ref_<function layer_norm_ref at 0x7830d1cbea20>] _
  iron/operators/gemv/test.py:251: in test_gemv_norm_prologue
      assert not errors, f"Test failed with errors: {errors}"
  E   AssertionError: Test failed with errors: {'output': [0]}
  E   assert not {'output': [0]}
  ----------------------------- Captured stdout call -----------------------------
  Device: <abc.NPU2 object at 0x7830d1731400>
  Matrix dimensions: M=2048, K=4096
  Tiling: m_input=1, m_output=2048
  Columns: 1
  Mismatch in output[0]: expected -310.000000, got -372.000000

```

PR #177, opened 2026-08-31
gemm: software-pipelined k-loop + INT8 dtype support

bong-water-water-bong:
## Summary
Software-pipeline the k-loop in `matmul_vectorized_2x2_mmul` and add validated INT8/INT16 dtype support to the GEMM operator.

### Kernel (aie_kernels/aie2p/mm.cc)
- Software-pipelined k-loop: prologue loads k-step 0, loop body consumes the previous step's A/B tiles while prefetching the next, epilogue MACs the final step — hides load→vmac latency on AIE2P.
- Pipelining hint clamped to the real compile-time trip count (colA−1, in [1,4]) so small K tiles are not mis-scheduled.

### Operator (op.py, design.py)
- Per-dtype tile-multiple validation matching kernel static_asserts.
- `_kernel_dtype_flag` selects the correct `{combo}_ONLY` define; `bf16_f32_ONLY` under `prio_accuracy`.
- Integer outputs narrower than the 32-bit accumulator rejected (i8→i8/i8→i16/i16→i16 truncate).
- design.py: duplicate npu1 entry merged; i8/i16 MAC dims.

### Validation (NPU Strix Halo, XDNA2)
- i8→i32 bit-exact vs numpy int32 reference at 1024×512×1024, 2048×512×2048, 2048×2048×2048, 2048×2048×8192 (N-partition).
- ~8.4 TOPS best at 2048³.


bong-water-water-bong:
Closing — superseded. The INT8 + software-pipelined-k-loop work landed on devel in refined form (`6410550` → `0f5cc48` → `5582ca1`), plus the llama prefill fusion (`76eee2a`) and real-prompt-length sizing (`764e7ef`) that build on it. The feature branch's older iterations (e670793/84c276a) are fully contained in those devel commits; keeping this PR open would conflict with the same files. Work is verified on Strix Halo XDNA2 (bit-exact INT8, deterministic prefill).

PR #178, opened 2026-08-31
Enable trace in OperatorSequence

andrej:
Requires upstream MLIR-AIE PR https://github.com/Xilinx/mlir-aie/pull/3652

Supersedes #154


andrej:
This PR requires https://github.com/Xilinx/mlir-aie/pull/3652 (multi aiex.run-trace-enablement), but that PR landed after https://github.com/Xilinx/mlir-aie/pull/3553 (objectFifo refactor), which is a syntax-breaking change that removes the `objectfifo.subview` op. Stream still emits `objectfifo.subview`s, cc @asyms. Stream will therefore need to update and remove the usage of `objectfifo.subview`. I will temporarily disable Stream testing so we can get this in; opened issue #194 for this. The tracing functionality is tested and works. Sorry for the complications.

hunhoffe (approved):

PR #179, opened 2026-08-31
operators: propagate dtype into AIERuntimeArgSpec

atassis:
## Problem

`AIERuntimeArgSpec.dtype` defaults to `bfloat16`, and none of the three `get_arg_spec()` implementations pass `dtype=`. So the spec reports bf16 regardless of what the design actually moves.

Concretely, `GEMM(dtype_in='i8', dtype_out='i32')` reports its C arg spec as bfloat16 -- **32768 bytes** -- while the design DMAs int32, **65536 bytes**. A caller sizing a host buffer from the spec under-allocates by 2x. The same 2x applies to `StridedCopy` and `Repeat` with a non-default dtype.

Two in-tree consumers size buffers off that value: `test_utils.py`'s `XRTTensor` sizing and `sequence.py`'s `calculate_buffer_layout`.

The defect is **latent today** -- a grep of every constructor call site in `iron/` shows nothing currently passes a non-default dtype -- so this is a landmine rather than a live failure.

## Fix

Pass the dtype through at the 7 call sites across 3 operator files: `dtype=self.dtype` for `strided_copy` and `repeat`, `dtype=str_to_dtype(self.dtype_in/out)` for `gemm`.

The two supporting commits are what the new test needs: `comparison.py` gains a strict comparison so exact equality is expressible, and `conftest.py` no longer aborts the whole session on an unparametrized test.

## Test

New `iron/tests/common/arg_spec_dtype.py`, 6 device-free cases.

Verified red-then-green by hand rather than only after: **5 of 6 fail before the change, 6 of 6 pass after.** (Run by importing the module and calling the test functions directly -- `conftest.py` resolves the device at collection time, so even `pytest --collect-only` would open the NPU.)

Both consumers confirmed to compute correct byte counts after the fix, and the change is a byte-for-byte no-op on every in-tree caller's default-dtype path.

`black` clean.

## Known limitation

`pr/strided-copy-transfer-size` also rewrites `get_arg_spec()` for an unrelated shape bug. Whichever lands second will need a small reconcile.

### DIFF
--- a/iron/operators/gemm/op.py
+++ b/iron/operators/gemm/op.py
@@ -15,6 +15,7 @@ from iron.common import (
     DesignGenerator,
 )
 from iron.common.device_utils import get_kernel_dir
+from aie.iron import str_to_dtype
 import aie.utils as aie_utils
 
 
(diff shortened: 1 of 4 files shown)


andrej (approved):

andrej:
Thanks!

PR #182, opened 2026-09-01
gemm: asymmetric INT8xINT4 (4-bit weights) on AIE2P 4x16x16 mmul — bit-exact + W4A8 llama decode

bong-water-water-bong:
## Summary
Adds asymmetric INT4 GEMM (A=i8, B=i4 packed two-per-byte) to the AIE2P 4x16x16 mmul path, plus the W4A8 llama inference arc that uses it.

## The core fix (issue: sizeof(int4) pointer arithmetic)
The AIE API's `int4_t` is an empty struct — `sizeof(int4) == 1` although each element is 4 bits. The kernel's manual `const int4*` pointer arithmetic (the j-block offset and the k-loop B advance) therefore moved 2x the real byte distance: the k-loop read k-blocks {0,2} then out-of-range, capping the effective K at half the tile. Fixed with `B_ADV = size_B / 2` for int4 (identity for int8) applied to every B pointer advance in `mm.cc` (`matmul_vectorized_2x2_mmul`).

## Verified
- INT4 GEMM bit-exact vs numpy int32 reference at all tested shapes (K=32/64/128/256, M up to 1024, N up to 8192 via partition_N) — 10/10 runs each.
- INT4 throughput 14.4 TOPS vs INT8 10.9 TOPS (1024x1024x4096) — 1.32x from the 4-bit density (the 4x16x16 mac is an emulated mixed-precision sequence on AIE2P, which is why it is not the full 2x).
- W4A8 llama-3.2-1B decode arc (subsequent commits): group-wise i4 scales, real C coherence fixes, 3.5x decode speedup, full README with repro commands.

## Commits
The PR spans the devel series from the asymmetric INT4 mmul (`0fef28b`) through the W4A8 llama decode + docs (`57853b8`). Happy to split if preferred.


llama_w4a8_npu.py
```diff
@@ -0,0 +1,389 @@
+#!/usr/bin/env python3
+"""Full-model W4A8 llama prefill on XDNA2 (part 30 follow-up).
```
andrej, line 2:
```suggestion
"""Full-model W4A8 llama prefill on XDNA2
```

Please instruct your AI to cleanse the rest of the changeset from similar internal/historical comments, thanks

andrej (commented):
This is great!

Do you have any performance numbers on the llama implementation using this?

In an effort to remove code duplication, we are in the process of migrating the C/C++ kernels to MLIR-AIE. @efurst is working on a PR that will remove all kernels from this repository. Could you file the parts of this that are changes to kernels to upstream MLIR-AIE, please? Once merged there, we can update the IRON pin of MLIR-AIE to a dev wheel that includes them and use this PR to incorporate the non-kernel changes here.

iron/common/sequence.py
```diff
@@ -579,8 +579,24 @@ def __init__(self, op, device_name="main", sequence_name="sequence"):
...
+            try:
+                xrt_context = pyxrt.hw_context(
+                    _rt._device, pyxrt.elf(str(op.artifacts[0].filename))
+                )
+                break
+            except RuntimeError as _e:
+                if _attempt == 7:
+                    raise
+                time.sleep(0.5)
+        assert xrt_context is not None
```
andrej, line 599: Please separate out this logic into its own PR and/or file against upstream MLIR-AIE DefaultNpuRuntime.

iron/applications/llama_3.2_1b/llama_cpu.py
```diff
@@ -293,6 +320,10 @@ def main():
     args = harness.parse_args()
     prompt = harness.get_prompt(args.prompt_len)
     config, state = harness.init(args.weights_path, args.tokenizer_path, prompt=prompt)
+    # DEBUG: cast weights to fp32 to measure bf16 fragility (LLAMA_FP32=1).
+    if __import__("os").environ.get("LLAMA_FP32"):
+        config.weights = {k: v.float() for k, v in config.weights.items()}
+        print("[cpu] weights cast to fp32", flush=True)
```
andrej, line 326: Inline `__import__("os")` deserves some cleanup. 

Maybe this debug infrastructure can be dropped altogether?

iron/operators/gemm/op.py
```diff
@@ -20,7 +20,22 @@
...
+    are rejected). ``dtype_in="bf16"`` with ``"bf16"`` or ``"f32"`` output
+    is the default floating-point path.
+
+    Known flake (XRT/amdxdna, not this operator): on NPU2 (Strix Halo) a
+    dispatch can rarely (~5% per process) return a wrong result after
+    several *distinct* xclbins have been compiled in one process — the
+    zero/accumulate write races the first submit on a freshly-registered
+    context. It self-heals on the next dispatch. The repo test harness
+    already warms up and verifies; production callers should do the same
+    (warm-up once, verify the first result, retry once on mismatch).
```
andrej, line 37: Do we know the root cause of this? This seems concerning -- if there is a data race, have you tried adding "emit_token=True" to _all_ tasks in the runtime sequence and awaiting each of them?

bench_int8_gemm.py
andrej, line 1: This probably belongs somewhere in a test subdirectory?

llama_w4a8_npu.py
andrej, line 1: This probably belongs in directory applications/llama_3.2_1b?

W4A8_LLAMA.md
andrej, line 1: --> iron/applications/llama_3.2_1b

llama_w4a8_validate.py
```diff
@@ -0,0 +1,87 @@
...
+
+Usage (iron-venv python — has torch/safetensors + ml_dtypes + mlir_aie):
+  cd ~/amd-oss/iron
+  PYTHONPATH=/usr/lib/python3/dist-packages \
+    ~/amd-oss/iron-venv/bin/python llama_w4a8_validate.py [M] [layer]
+"""
+import sys
+import numpy as np
+
+sys.path.insert(0, "/home/bcloud/amd-oss/iron")
```
andrej, line 18: Pelase fix hard coded paths

llama_w4a8_npu.py
```diff
@@ -0,0 +1,389 @@
...
+  PYTHONPATH=/usr/lib/python3/dist-packages \
+    ~/amd-oss/iron-venv/bin/python llama_w4a8_npu.py "The capital of France is"
+"""
+import os
+import sys
+import math
+import time
+
+sys.path.insert(0, "/home/bcloud/amd-oss/iron")
+sys.path.insert(0, "/home/bcloud/amd-oss/iron/iron/applications/llama_3.2_1b")
```
andrej, line 25: fix hard-coded paths; adding to sys path should not be necessary in a properly set-up venv with installed IRON, so probably remove this altogether

andrej (changes requested):
I think the llama_w4a8_npu implementation needs the most work. Making sure it is tested in CI would be important, and I'd love to see it reuse some of the harness infrastructure that's already there for the bf16 llama implementation. 

Maybe this can be factored out into a separate PR so we can merge the INT8xINT4 GEMM separately first?

PR #183, opened 2026-09-01
ci: guard the PR comment against GitHub's 65536-character limit

atassis:
`pr-comment.yml` posts the aggregate body in one call. Once all four suites have reported for a head SHA it can exceed what GitHub accepts: on run 33550798970 the body was 313,521 characters and the API returned `422 Validation Failed: body is too long (maximum is 65536 characters)`, so the PR received no results comment. Earlier runs on the same SHA posted fine because fewer suites had finished.

The 65536 in that message is not the enforced ceiling. Every aggregate comment on #150-#183 runs between 122,494 and 142,053 characters and all of them posted, so a guard comparing against a constant would strip the details off comments GitHub accepts today. Post the full body instead, and shed a section only when GitHub actually rejects one.

Sections go in size order. The trend tables are the bulk of an oversized body, so they go first: on run 33550798970 that takes the body from 313,521 to 83,961 characters and keeps every per-suite pass/fail table. After that the per-suite details, then the summary halved until it fits. The marker stays in the body on every rung, so the update branch keeps finding the existing comment instead of creating a new one each run.

Verified by replaying that run's four result artifacts through the script with a stubbed API rejecting above a given size: at 262144 it lands on the trends-dropped rung, at 65536 on the summary, and with nothing over the limit it makes a single call with the body unchanged.

## Added
- Fallback ladder in the `Comment on PR` step, with a `core.warning` when the comment had to be shortened.

## Changed
- Body assembled from a `header` constant so each rung can be rebuilt from it.
- Create and update go through one `post()` helper.

## Removed
-

## PR Merge Checklist

### DIFF
--- a/.github/workflows/pr-comment.yml
+++ b/.github/workflows/pr-comment.yml
@@ -217,14 +217,31 @@ jobs:
             }
 
             const marker = '<!-- iron-ci-aggregate -->';
-            const body = `${marker}
+            const header = `${marker}
             ## CI Test Results
 
             [${commitSha}](${server}/${repo}/commit/${commitSha}) (${date})
 
-            ${summaryContent}
+            `;
 
-            ${details}`;
+            // GitHub rejects an issue comment body over 65536 characters. A run
+            // where most tests fail produces a per-suite table several times
+            // that, so shed the details and then the summary rather than lose
+            // the comment entirely (422 Validation Failed, no results posted).
+            const LIMIT = 65536;
+            const runUrl = `${server}/${repo}/actions/runs/${context.payload.workflow_run.id}`;
+            const elided = `\n_Truncated: the full report exceeds GitHub's ${LIMIT}-character comment limit. See the [workflow run](${runUrl}) for complete results._\n`;
+
+            let body = header + summaryContent + '\n\n' + details;
+            if (body.length > LIMIT) {
+              body = header + summaryContent + elided;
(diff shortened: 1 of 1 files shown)


andrej (approved):

andrej:
Any good way to test this? Maybe we could make a lot of tests fail on purpose in this PR so we can see it in action (and we can fix them before we merge for real)?

atassis:
Testing it from this branch won't work: `pr-comment.yml` is `workflow_run`-triggered, so GitHub always runs devel's copy. This PR actually demonstrated that itself, its own comment runs (33567717223, 33568280474) hit the same 422.

Good news is we don't need broken tests. The failure reproduces on any SHA where all four suites have reported, a279467 does it every time.

Bad news is you shouldn't merge this yet. I checked the comments on #150-#183: they run 122k to 142k characters and all of them posted fine, so 65536 isn't the limit GitHub actually enforces. As written my guard would strip the details off every comment, not just the huge ones.

Let me redo it to shed only on a real 422, and drop `trends.md` first, that keeps the pass/fail tables. I'll ping you when it's up.

atassis:
Pushed as fe71df5. It now sheds only when the API actually rejects the body, trend tables first, which on the run that failed leaves an 84k comment with every pass/fail table intact. Body updated too.

andrej (approved):

PR #185, opened 2026-09-06
compilation: stop pinning aiecc to one thread

atassis:
## Problem

Both aiecc rules default `AIECC_JOBS` to `1`:

```python
options = [f"-j{os.environ.get('AIECC_JOBS', '1')}"]
```

so every design's per-core compiles run one at a time, whatever the machine has. aiecc's own default
is auto-detect, and nothing here explains the override.

## Fix

Default to `0`, which is aiecc's auto-detect. The environment variable still overrides, so anyone who
wants one thread keeps it.

## Test

On a 24-core design (`StaticMHA`, 20 heads, 8 pipelines): aiecc goes from **7.7 s to 6.0 s**. Nothing
above 8 threads helps.

`-j` does not change what aiecc produces, so this is safe rather than a trade: `insts.bin` and all 24
per-core ELFs are **byte-identical between `-j1` and `-j16`**, and `input_with_addresses.mlir` differs
only in the work-dir path it embeds — which two runs at the *same* `-j` also differ in.

### DIFF
--- a/iron/common/compilation/base.py
+++ b/iron/common/compilation/base.py
@@ -556,6 +556,15 @@ def _link_build_outputs_into(work_dir: Path, build_dir: Path) -> None:
     link_files_from(build_dir / get_kernel_dir())
 
 
+# aiecc's own default. "1" here made every design's per-core compiles serial: on
+# the encoder-MHA design (24 cores) aiecc costs 7.7 s at -j1 and 6.0 s at -j0, and
+# nothing above 8 helps. Safe because -j does not change what aiecc produces --
+# measured on that design, insts.bin and all 24 per-core ELFs are byte-identical
+# between -j1 and -j16, and input_with_addresses.mlir differs only in the work-dir
+# path it embeds, which two runs at the SAME -j also differ in.
+_AIECC_DEFAULT_JOBS = "0"
+
+
 class AieccCompilationRule(CompilationRule):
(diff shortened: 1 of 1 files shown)


iron/common/compilation/base.py
```diff
@@ -556,6 +556,15 @@ def link_files_from(directory: Path) -> None:
     link_files_from(build_dir / get_kernel_dir())
 
 
+# aiecc's own default. "1" here made every design's per-core compiles serial: on
+# the encoder-MHA design (24 cores) aiecc costs 7.7 s at -j1 and 6.0 s at -j0, and
+# nothing above 8 helps. Safe because -j does not change what aiecc produces --
+# measured on that design, insts.bin and all 24 per-core ELFs are byte-identical
+# between -j1 and -j16, and input_with_addresses.mlir differs only in the work-dir
+# path it embeds, which two runs at the SAME -j also differ in.
```
andrej, line 564:
```suggestion
# Default parallelism setting used when environment variable AIECC_JOBS is not set.
# 0 means auto-detect.
```

andrej (approved):
Is anything setting `AIECC_JOBS = 1` for CI? If not, please add that, our our CI will constantly OOM.

andrej:
@atassis Can you please pin AIECC_JOBS=1 in the CI scripts so I can merge this?

atassis:
@andrej Hi
Done

PR #186, opened 2026-09-06
compilation: run a rule's independent commands concurrently

atassis:
A compilation rule whose commands are independent of one another now has them run
concurrently instead of one after the next.

## Added
- `commands_are_independent` on `CompilationRule`, default off. A rule sets it when its
  `compile()` emits one command per worklist artifact with none of them consuming
  another's output.
- `IRON_COMPILE_JOBS` to override the concurrency.

## Changed
- `_execute_rule` runs a rule's commands in parallel when the rule declares them
  independent, and serially otherwise. Kernel object compiles are the case that motivated
  it: they were costing the sum of their walls, 7.9 s of a 17.1 s encoder-MHA build for two
  kernels. Bounded by cores — each kernel compile is a single-threaded clang peaking near
  205 MB of RSS.

## Removed
- Nothing.

## PR Merge Checklist

### DIFF
--- a/iron/common/compilation/base.py
+++ b/iron/common/compilation/base.py
@@ -41,6 +41,7 @@ import os.path
 import shutil
 import zlib
 import logging
+import concurrent.futures
 import subprocess
 import importlib.util
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


iron/common/compilation/base.py
```diff
@@ -103,14 +104,46 @@ def plan(
...
+def _execute_rule(rule: CompilationRule, commands: list[CompilationCommand]) -> None:
+    """Run one rule's commands, concurrently when the rule says they are independent.
+
+    A rule that sets `commands_are_independent` emits one command per artifact in
+    its worklist, with no artifact depending on another's output -- kernel object
+    compiles are the case that matters, and they were costing the SUM of their
+    walls (7.9 s of a 17.1 s encoder-MHA build for two kernels).
+
+    Bounded by cores: each kernel compile is a single-threaded clang peaking near
+    205 MB of RSS. Set IRON_COMPILE_JOBS to override.
```
andrej:
```suggestion

    A rule that sets `commands_are_independent` emits one command per artifact in
    executes them in parallel. Set IRON_COMPILE_JOBS to override.
```

iron/common/compilation/base.py
```diff
@@ -471,6 +504,12 @@ def __repr__(self) -> str:
 class CompilationRule(ABC):
     """A compilation rule is applied to a artifact graph, producing compilation commands and a transformed artifact graph."""
 
+    #: Set by a rule whose `compile()` emits one command per worklist artifact,
+    #: none of them consuming another's output. `execute` may then run them
+    #: concurrently. Default off: a rule that batches dependent steps into one
+    #: application must keep its order, and only the rule knows which it is.
+    commands_are_independent: bool = False
```
andrej, line 510:
```suggestion
    # Set by a rule whose `compile()` emits one command per worklist artifact,
    # none of them consuming another's output. `execute` may then run them
    # concurrently. Default off: a rule that batches dependent steps into one
    # application must keep its order, and only the rule knows which it is.
```

andrej (approved):
As with the other PR, please add something that disables this on CI, or we'll likely OOM. Then I think this is good. Thanks!

andrej:
@atassis Looks like there's currently tests failing. I'm closing this for now since I haven't seen recent activity on here. This doesn't mean the PR is rejected -- if you choose to pick it up again, do not hesitate to reopen. Just closing to keep things organized so I don't miss things (a reopened PR will be easier for me to spot).

And just so you know, long-term, we will replace the compile infrastructure in IRON by reusing the CompileableDesign infrastructure in MLIR-AIE.

PR #187, opened 2026-09-08
Port aie_kernels to compile under xchesscc (compiler="chess")

bong-water-water-bong:
Ports the entire `aie_kernels` C++ library to compile under `xchesscc` so every operator builds with `compiler="chess"` and passes its reference test, without regressing the Peano path.

## Toolchain (2024 vitis_aie_essentials recipe — CHESS_PORT_STATE.md has full context)
- xchesscc from Ryzen AI 1.3.0 `vitis_aie_essentials` (unguarded acquire; 2025.2+/2026.1 chess emits the guarded acquire that never completes on NPU2 silicon — upstream Xilinx/mlir-aie#3690)
- patched mlir-aie v1.4.2 aiecc: bare peanoElfs link (`AIE_AIECC_NO_XBRIDGE=1`), me_primitive stubs, self-contained aie2p softfloat, `fixChessFloatLiterals` (.0 suffix for integer-valued f32 constants), archive-order fix for fused OperatorSequence links
- harness: `--compiler chess` option (conftest + `iron/common/compilation`)

## Kernel fixes (all peano-safe)
- bf16 scalar call-args → float + `(bfloat16)` casts (chess LLVM rejects bf16 call constants): leaky_relu, mha `partial_softmax` inv_scale chain
- softmax: explicit `(float)` casts for comparison ambiguities
- `aie2/softmax.cc`: `-INFINITY` → `(bfloat16)(-65504.0f)` (aie2 chess front-end headers lack INFINITY)
- rounding-mode fix in `generic/mul.cc` (RNE for bf16 conversion)

## Test evidence (compiler="chess", 2024 essentials, NPU2)
- axpy 20/20 + silu 20/20 (bare link)
- 11 elementwise/activation ops: 400/400
- gemm 45 / gemv 95 / dequant 40 / mha 5 = 185/185
- transpose/strided_copy/repeat/mem_copy 140/140 + axpy 20/20
- Qwen3-0.6B single-layer smoke: LAYER_OK + HEAD_OK under chess and peano
- rope/swiglu reference tests: chess 55 passed, peano 55 passed
- peano regression clean across all suites; clean-state re-runs verified

### DIFF
--- a/aie_kernels/aie_kernel_utils.h
+++ b/aie_kernels/aie_kernel_utils.h
@@ -69,4 +69,26 @@
 #define AIE_LOOP_FLATTEN
 #endif
 
+// Runs `body` (a zero-arg lambda) `count` times. When count >= MinIters the loop is
+// eligible for software pipelining (extra pragma macros may be passed after `body`,
+// e.g. AIE_PREPARE_FOR_POSTPIPELINING); otherwise a plain no-unroll loop is emitted,
+// avoiding invalid pipeliner assumptions for tiny trip counts.
+#define VERSIONED_LOOP(MinIters, count, body, ...)                                                                     \
+    do {                                                                                                               \
+        if ((count) >= (MinIters)) {                                                                                   \
+            __VA_ARGS__                                                                                                \
+            AIE_LOOP_RANGE(MinIters, )                                                                                 \
+            for (int _vl_i = 0; _vl_i < (count); _vl_i++) {                                                            \
+                (body)();                                                                                              \
+            }                                                                                                          \
+        } else {                                                                                                       \
+            AIE_NO_PREPARE_FOR_PIPELINING                                                                              \
+            AIE_LOOP_RANGE(1, )                                                                                        \
+            AIE_LOOP_NO_UNROLL                                                                                         \
+            for (int _vl_i = 0; _vl_i < (count); _vl_i++) {                                                            \
+                (body)();                                                                                              \
+            }                                                                                                          \
+        }                                                                                                              \
+    } while (0)
+
 #endif
--- a/.gitignore
+++ b/.gitignore
@@ -1,24 +1,2 @@
-# SPDX-FileCopyrightText: Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
-# SPDX-License-Identifier: Apache-2.0
-
-/my_install*
-/ironenv*
-.vscode
-__pycache__
-.DS_Store
-*.bin
-build/*
-**/_build/**
-**/build/**
-**/build_elf/**
-*.exe
-*.csv
-secret_github_token
-id_ed25519
-id_ed25519.pub
-*.log
-*.safetensors
-*.model
-.cline_storage
-*.egg-info
-**/*.prj/**
+# Created by venv; see https://docs.python.org/3/library/venv.html
+*
--- a/CHESS_PORT_STATE.md
+++ b/CHESS_PORT_STATE.md
@@ -1,0 +1,518 @@
+# Chess port state (2026-09-07) — strixhalo
+Goal: compile IRON aie_kernels with compiler="chess" (task-1: axpy).
+
+## Installed chess toolchains
+- Vitis 2025.2: /home/bcloud/Xilinx2025/2025.2/Vitis/aietools (NEW, 75GB, minimal config)
+  - arch for AIE2P-class = **aie2ps** (no aie2p). Symlink alias added:
+    data/aie2p -> aie2ps ; tps/lnx64/target_aie2p -> target_aie2ps
+- Vitis 2026.1: /home/bcloud/Xilinx/2026.1/Vitis/aietools (has aie2p + aie2ps; model renamed acquire to *guarded*)
+- Ryzen AI essentials 1.3.0 (2024): ~/Downloads/ryzen_ai-1.3.0/vitis_aie_essentials (arch aie2p, UNGUARDED void_acquire)
+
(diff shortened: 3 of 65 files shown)


hunhoffe:
Hello @bong-water-water-bong ! For the most part, we're migrating in the direction of more peano support and less chess support https://github.com/Xilinx/mlir-aie/discussions/3414 

Is there a reason that you prefer chess at this time?

bong-water-water-bong:
Hi @hunhoffe — thanks for the question. Short version: we don't prefer chess as a direction; peano is our default too, and this PR keeps it that way — the default path is untouched, every suite is re-run clean under peano, and none of the kernel changes are chess-gated. We invested in making `compiler="chess"` actually *work* because of what sits behind the chess path, not because we prefer the compiler:

1. **aiesim is the real anchor.** As the #3414 thread itself concludes, the piece that is genuinely chess-only today is the cycle-accurate simulator, not the compiler. We do cycle-accurate bring-up of new operators before device runs, and that path requires kernels to compile under xchesscc. Until the ISS has an open replacement, a working chess compile route is what keeps hardware-free, cycle-accurate verification possible — including for people who can't occupy scarce device runners.

2. **Compiler-independent correctness evidence.** Running the same operator suite under both front-ends on the same NPU2 is a strong cross-check, and it surfaces toolchain bugs that look like kernel bugs when you only build with peano. This port is what exposed the 2025.2+/2026.1 chess **guarded-acquire** codegen that never completes on AIE2P silicon (Xilinx/mlir-aie#3690), and let us prove the workaround (2024 vitis_aie_essentials chess, unguarded acquire): chess-built kernels now compile, load, execute and pass reference tests on NPU2, alongside a clean peano regression.

3. **The kernel diffs are portability fixes, not chess-isms.** bf16 scalar call-constants, missing `INFINITY` in the aie2 chess headers, float/bfloat16 comparison ambiguity, exact float literals — all peano-safe (peano regression clean across every suite). They harden the kernel C++ against any toolchain, peano included, which seems aligned with the migration rather than against it.

So the intended shape is the one #3414 recommends: peano as default, chess as a maintained opt-in during the transition — not a reason to hold the migration. Happy to gate the chess coverage behind a marker or a separate CI job if you'd rather keep it out of the default path. I can also rebase onto current devel (the branch is behind and now conflicts). If chess support is being deprecated outright, tell me the timeline and I'll plan around it — but while aiesim and existing Vitis flows still need xchesscc-built kernels, letting the path rot silent-fails exactly the way it did before we fixed the toolchain pairing.

aie_kernels/aie_kernel_utils.h
andrej, line 78: I understand this helps with generality, but I wonder if with our limited program memory we will run into issues duplicating the loop body?

Might be good to at least provide an option of the kernels to the users where the user can assert, at compile-time, this _will_ run for `MinIters`, so don't emit the non-pipelined variant.

I'm also not sure if "VERSIONED_LOOP" is the clearest name?

andrej (commented):
Overall, I think making all kernels compil_able_ with both chess and peano is good. I'd want to keep a close eye on performance regressions for this one. 

I think widening the vector sizes from 16 --> 32 is fine as well and might get us a perf bump, though I'm not sure it's needed for this PRs stated goal of enabling chess? We should also double-check all call sites indeed call kernels with inputs that are multiples of 32 in size.

In an effort to reduce code duplication, we are currently migrating all kernels to MLIR-AIE. @efurst is working on a PR that will drop all kernels from this repository, so any changes here would be lost. Can I ask you to re-open this against upstream MLIR-AIE and close it here? Thank you, and sorry for that inconvenience -- long therm hopefully it'll benefit everyone, as any kernel improvements will be available everywhere. When you have the upstream PR open, please link it here as well, thanks.

bong-water-water-bong:
Hi @andrej — thanks for the clear steer, and understood on the MLIR-AIE migration direction (the aie_kernels dedup landing upstream in Xilinx/mlir-aie#3599 makes the rationale obvious).

Plan on our side, per your request:
1. Re-file this work against the upstream Xilinx/mlir-aie `aie_kernels` (the chess-compilable kernel changes are all peano-safe, so they carry over cleanly as compile-enablement), rather than IRON's legacy copies that are about to be dropped.
2. Link the upstream PR here once it's open.
3. Close this PR at that point.

One data point to weigh while we do that: upstream mlir-aie CI exercises the chess path compile-only today (no full chess ELF link, no NPU2 execution), and per Xilinx/mlir-aie#3690 the guarded-acquire chess cores never complete on NPU2 silicon — the working route is the 2024-essentials unguarded toolchain. So the upstream PR will carry the kernel-level chess enablement + fixes; silicon/runtime proof stays on the 2024-essentials recipe until #3690 is resolved.

Also happy to coordinate around @efurst's drop-kernels PR — if it lands first we'll rebase the upstream contribution accordingly.

andrej:
@bong-water-water-bong Thanks for your understanding. As an update, #203 is getting close to ready to merge and the kernels are already in upstream MLIR-AIE.

> One data point to weigh while we do that: upstream mlir-aie CI exercises the chess path compile-only today (no full chess ELF link, no NPU2 execution), and per https://github.com/Xilinx/mlir-aie/issues/3690 the guarded-acquire chess cores never complete on NPU2 silicon — the working route is the 2024-essentials unguarded toolchain. So the upstream PR will carry the kernel-level chess enablement + fixes; silicon/runtime proof stays on the 2024-essentials recipe until #3690 is resolved.

I haven't read up on the specific issue. It sounds like enablement at compile-time only (this PR) is a good first step, and fixing the runtime issue can be part of the solution to issue #3690.

PR #191, opened 2026-09-08
rope: make the CPU reference match the device's angle-row convention

### LINKED ISSUE #188: [BUG] RoPE's CPU reference and device kernel disagree on angle-row mapping for 1 < angle_rows < rows
## Describe the bug

`iron/operators/rope/reference.py`'s `reference()` implements a different `angle_rows`-to-row mapping than the device kernel it's meant to check.

`design.py`'s `core_body` acquires one angle row and applies it to `rows / angle_rows` **consecutive** input rows before moving to the next angle row -- row `r` uses angle row `r // (rows / angle_rows)`. `reference()` instead does `cos = cos.repeat(rep, 1)` (`rep = rows // angle_rows`), which tiles the whole `angle_rows`-row block `rep` times -- row `r` uses angle row `r % angle_rows`. The two mappings agree only when `angle_rows` is 1 or `rows`; for any `1 < angle_rows < rows` they disagree on most rows, silently (no error, just a wrong expected value).

This is live-shaped, not a corner case invented for the report: `llama_3.2_1b/llama_npu.py`'s prefill RoPE calls are exactly this shape (`RoPE(rows=prompt_len*n_heads, angle_rows=prompt_len, ...)`). It hasn't fired because nothing currently calls `RoPE.reference()` at that shape -- `rope/test.py` builds its golden through `generate_golden_reference()`/`apply_rope()` instead (a different code path, indexed straight off the full cos/sin table, with an explicit `.transpose(0, 1)` to get the position-major layout right) and llama's prefill RoPE only uses the compiled callable, never `.reference()`. So this is one PR away from firing: anyone who calls `RoPE(...).reference(x, angles)` to check a batched (`1 < angle_rows < rows`) run gets a wrong expected answer with no error.

## To Reproduce

```python
import torch
from iron.operators.rope.reference import reference

rows, angle_rows, cols = 6, 3, 4
half = cols // 2
x = torch.randn(rows, cols).to(torch.bfloat16)
angles = torch.zeros(angle_rows, cols, dtype=torch.bfloat16)
angles[:, 0::2] = torch.rand(angle_rows, half).to(torch.bfloat16)
angles[:, 1::2] = torch.rand(angle_rows, half).to(torch.bfloat16)

# What the device kernel actually does (design.py's core_body): row r uses
# angle row r // (rows // angle_rows) -- consecutive rows share an angle row.
device_angle_idx = [r // (rows // angle_rows) for r in range(rows)]
print("device convention:   ", device_angle_idx)   # [0, 0, 1, 1, 2, 2]

# What reference()'s cos.repeat(rep, 1) actually selects per row.
cos = torch.arange(angle_rows).reshape(angle_rows, 1).expand(angle_rows, half).float()
ref_angle_idx = cos.repeat(rows // angle_rows, 1)[:, 0].long().tolist()
print("reference() convention:", ref_angle_idx)    # [0, 1, 2, 0, 1, 2]

# 4 of 6 rows use a different angle row -> reference(x, angles, rows=rows, cols=cols)
# disagrees with the device on 4 of 6 output rows.
```

Output on `devel` (`deb6e1e7cd1e3aefcf7530108a4647a1d967ce18`):
```
device convention:    [0, 0, 1, 1, 2, 2]
reference() convention: [0, 1, 2, 0, 1, 2]
```
Running the full `reference(x, angles, rows=rows, cols=cols)` and comparing row-by-row against a ground truth built from the device convention: 4 of 6 rows differ.

## Expected behavior

`reference()` should select the same angle row per input row as `design.py`'s `core_body`: consecutive rows, `r // (rows // angle_rows)`.

## Additional context

I'm fairly confident the device kernel is the intended semantics, not `reference()`: it matches `design.py`'s own docstring ("each row of angles will be reused for `rows / angle_rows` consecutive rows of the input tensor"), it's what `llama_npu.py`'s real prefill usage assumes, and `rope/test.py`'s own golden generation independently encodes the same consecutive-row convention via its `.transpose(0, 1)`. On that basis I've also prepared a PR fixing `reference()` (`cos.repeat` -> `cos.repeat_interleave`) rather than filing this as an open question -- happy to drop it if the intended convention is actually the other way around.

atassis:
Fixes #188.

`reference()` used `cos.repeat(rep, 1)` to stretch an `angle_rows`-row LUT to `rows` rows, which tiles the whole block (row `r` uses angle row `r % angle_rows`). The device kernel (`design.py`'s `core_body`) applies one angle row to `rows / angle_rows` *consecutive* input rows instead (row `r` uses angle row `r // (rows / angle_rows)`). The two conventions only agree at `angle_rows` in `{1, rows}`, so this was invisible until something calls `reference()` at `1 < angle_rows < rows` -- exactly the shape `llama_npu.py`'s prefill RoPE uses (`rows=prompt_len*n_heads, angle_rows=prompt_len`), though nothing currently routes through `reference()` at that shape. `repeat_interleave` is the direct fix: it repeats each row in place instead of tiling the block.

I'm recommending the device kernel as canonical over `reference()`, not the other way around -- see the linked issue for why.

## Added
- `iron/tests/operators/rope_reference_convention.py`: builds a ground truth directly from the device's row-to-angle mapping and checks `reference()` against it across several `(rows, angle_rows)` shapes, including the decisive `1 < angle_rows < rows` case.

## Changed
- `iron/operators/rope/reference.py`: `cos.repeat(rep, 1)` / `sin.repeat(rep, 1)` -> `repeat_interleave(rep, dim=0)`, and the docstring now says which convention and why.

## Removed
-

## Evidence
```
$ PYTHONPATH=. python3 -m pytest --noconftest iron/tests/operators/rope_reference_convention.py -v
collected 2 items

rope_reference_convention.py::test_reference_matches_device_convention_for_batched_angle_rows PASSED
rope_reference_convention.py::test_reference_matches_device_convention_across_shapes PASSED

======================== 2 passed in 0.89s ========================
```
Before the fix, the first test's `rows=6, angle_rows=3` case shows 4 of 6 rows mismatching the device-convention ground truth (see the linked issue for the exact numbers); after `repeat_interleave`, all shapes tested (`(6,3), (8,2), (1024,1), (4,4), (13,13), (12,4)`) match exactly.

Not touched: `generate_golden_reference()`/`apply_rope()`, which `rope/test.py` actually uses for its device-comparison golden -- that path already encodes the consecutive-row convention correctly via its `.transpose(0, 1)`, so it isn't affected by this bug and this PR doesn't change it.

## PR Merge Checklist

### DIFF
--- a/iron/tests/operators/rope_reference_convention.py
+++ b/iron/tests/operators/rope_reference_convention.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""reference() must apply an angle row to `rows / angle_rows` CONSECUTIVE
+input rows, matching the device kernel -- not to `rows / angle_rows` tiled
+copies of the whole angle block.
+
+design.py's core_body acquires one angle row and applies it to
+`tensor_rows_per_angle_row` consecutive input rows before moving to the
+next angle row: row r uses angle row `r // (rows // angle_rows)`. A prior
+version of reference() used `cos.repeat(rep, 1)`, which tiles the whole
+angle block `rep` times (row r uses angle row `r % angle_rows`) --  the
+interleaved convention. The two conventions agree only when angle_rows is
+1 or rows, so this is invisible unless something exercises
+1 < angle_rows < rows -- exactly the shape llama_npu.py's prefill RoPE uses
+(rows=prompt_len*n_heads, angle_rows=prompt_len), though nothing there
+currently calls reference() with it.
+"""
+
+import torch
+
+from iron.operators.rope.reference import reference
+
+
+def _block_major_expected(x, angles, rows, angle_rows):
+    """Ground truth built directly from the device convention: row r uses
+    angle row r // (rows // angle_rows)."""
+    cols = x.shape[-1]
+    half = cols // 2
+    tensor_rows_per_angle_row = rows // angle_rows
+    out = torch.empty(rows, cols, dtype=torch.float32)
+    for r in range(rows):
+        a = r // tensor_rows_per_angle_row
+        cos = angles[a, 0::2].to(torch.float32)
+        sin = angles[a, 1::2].to(torch.float32)
+        x1, x2 = x[r, :half].to(torch.float32), x[r, half:].to(torch.float32)
+        out[r, :half] = x1 * cos - x2 * sin
+        out[r, half:] = x2 * cos + x1 * sin
+    return out.to(torch.bfloat16)
+
+
+def _make_inputs(rows, angle_rows, cols=4, seed=0):
+    torch.manual_seed(seed)
+    half = cols // 2
+    x = torch.randn(rows, cols).to(torch.bfloat16)
+    angles = torch.zeros(angle_rows, cols, dtype=torch.bfloat16)
+    angles[:, 0::2] = torch.rand(angle_rows, half).to(torch.bfloat16)
+    angles[:, 1::2] = torch.rand(angle_rows, half).to(torch.bfloat16)
+    return x, angles
+
+
+def test_reference_matches_device_convention_for_batched_angle_rows():
+    """Decisive case: 1 < angle_rows < rows. Before the fix, 4 of these 6
+    rows disagreed with the device convention."""
+    rows, angle_rows = 6, 3
+    x, angles = _make_inputs(rows, angle_rows)
+    expected = _block_major_expected(x, angles, rows, angle_rows)
+    got = reference(x, angles, rows=rows, cols=x.shape[-1])
+    assert torch.equal(expected, got)
+
+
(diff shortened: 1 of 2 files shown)


iron/tests/operators/rope_reference_convention.py
```diff
@@ -0,0 +1,68 @@
...
+"""reference() must apply an angle row to `rows / angle_rows` CONSECUTIVE
+input rows, matching the device kernel -- not to `rows / angle_rows` tiled
+copies of the whole angle block.
+
+design.py's core_body acquires one angle row and applies it to
+`tensor_rows_per_angle_row` consecutive input rows before moving to the
+next angle row: row r uses angle row `r // (rows // angle_rows)`. A prior
+version of reference() used `cos.repeat(rep, 1)`, which tiles the whole
+angle block `rep` times (row r uses angle row `r % angle_rows`) --  the
+interleaved convention. The two conventions agree only when angle_rows is
```
andrej: A prior version

andrej (changes requested):

andrej (approved):

PR #192, opened 2026-09-08
conftest: only resolve the device when a collected test restricts to one

atassis:
Root `conftest.py`'s `pytest_collection_modifyitems` calls `aie_utils.DefaultNPURuntime.device()` unconditionally, for every collection, regardless of whether any collected test actually needs a specific device. `pytest.ini`'s `testpaths` spans `iron/operators iron/applications iron/tests` with no guard, so a plain `pytest` anywhere in this tree opens the NPU -- contending with anything else already using the (single-tenant) device, and failing outright on a machine with none attached.

This is the same defect three recent merged PRs removed from other code paths for the same reason -- `get_current_device()` exists specifically for "offline inspection paths that must not initialize the runtime" (#184: bind MHA compile to it instead of `DefaultNPURuntime`; #165: make `OperatorSequence` compilable without XRT; #164: same for `torch`). This hook is the one site still doing the eager version, for a purpose (skipping device-restricted tests) that doesn't need it when nothing in the collected set is device-restricted.

## Added
- `iron/tests/infrastructure/conftest_lazy_device.py`: loads the real root `conftest.py` by path and calls its hook directly with fake items and a stubbed `DefaultNPURuntime` that raises if `.device()` is called, confirming it's skipped when nothing carries `@pytest.mark.supported_devices` and that the marker-skip logic is unchanged when something does.

## Changed
- `conftest.py`: `pytest_collection_modifyitems` only calls `DefaultNPURuntime.device()` when at least one collected item carries `@pytest.mark.supported_devices`.

## Removed
-

## Evidence
```
$ PYTHONPATH=. python3 -m pytest --noconftest iron/tests/infrastructure/conftest_lazy_device.py -v
collected 2 items

conftest_lazy_device.py::test_no_device_probe_when_nothing_is_device_restricted PASSED
conftest_lazy_device.py::test_device_probed_and_marker_logic_preserved_when_a_test_is_restricted PASSED

======================== 2 passed in 0.34s ========================
```
Ran with `--noconftest` since the module under test is loaded directly by path rather than through pytest's own conftest machinery. Confirmed the first test fails against unmodified `devel` (`deb6e1e7`) with the stub's `_DeviceCalledError`, i.e. `.device()` really is called today with zero device-restricted tests collected.

## PR Merge Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -175,10 +175,20 @@ def pytest_configure(config):
 
 
 def pytest_collection_modifyitems(config, items):
+    marked_items = [
+        (item, item.get_closest_marker("supported_devices")) for item in items
+    ]
+    marked_items = [(item, marker) for item, marker in marked_items if marker]
+    if not marked_items:
+        # No collected test restricts itself to specific devices, so nothing
+        # here needs the NPU. Resolving it anyway made a plain `pytest`
+        # collection open the device unconditionally, contending with
+        # whatever else is using it and failing outright with no NPU at all.
+        return
+
     device = aie_utils.DefaultNPURuntime.device().resolve().name
-    for item in items:
-        marker = item.get_closest_marker("supported_devices")
-        if marker and device not in marker.args:
+    for item, marker in marked_items:
+        if device not in marker.args:
             item.add_marker(
                 pytest.mark.skip(
                     reason=f"Not supported on {device} (supported: {', '.join(marker.args)})"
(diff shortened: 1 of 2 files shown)


andrej (changes requested):
history narrating comments; please fix this in all future PRs before submitting

andrej (approved):

PR #198, opened 2026-09-11
kernels: five elementwise loops walked off the end of their tiles

atassis:
Five elementwise kernels bounded their vector loop on the tile size rather than
on the last full vector. `load_v`/`store_v` are full width regardless of how many
elements remain, so a tile that does not divide the vector width either leaves
the tail unwritten or reads and writes past the buffer.

```
add.cc          F = size/32, i < F        tail never written; consumer reads stale
mul.cc          i < size, i += 32         writes up to 31 elements past c
axpy.cc         i < vector_size, i += 64  writes up to 63 past x, y, z
convert_copy.cc i < vector_size, i += 16  writes up to 15 past both
rope.cc         v < dims_half, v += N     reads/writes past input, lut, output
```

In L1 the memory an overrun lands on is another objectFIFO buffer or the stack,
so this is silent rather than a crash. It stays hidden while every tile divides
its kernel's vector width, which is the common case; a tile of 80 against a
32-wide vector (80 = 32*2 + 16) leaves the last 16 elements of every output stale.

Each kernel gets a scalar tail rather than a divisibility check on the operator,
because a check would reject shapes that are handled correctly today. Where the
tile already divides the vector width the tail is dead code, so shapes that work
now are unaffected.

## Added
- Scalar tail loops in `add.cc`, `axpy.cc`, `convert_copy.cc`, `mul.cc`, `rope.cc`.

## Changed
- The vector loops in those kernels stop at the last full vector.

## Removed
-

## PR Merge Checklist

### DIFF
--- a/aie_kernels/generic/add.cc
+++ b/aie_kernels/generic/add.cc
@@ -38,6 +38,15 @@ template <typename T_in, typename T_out> void eltwise_vadd(T_in *a, T_in *b, T_o
         aie::store_v(pC1, cout);
         pC1 += vec_factor;
     }
+    // `size` is a caller-chosen tile and need not divide vec_factor, so the vector body is
+    // bounded on the last FULL vector and the remainder is handled scalar-wise here. load_v and
+    // store_v are full-width regardless of how many elements remain, so a short final vector
+    // iteration would read and write past the buffer; in L1 that is another objectFIFO buffer or
+    // the stack.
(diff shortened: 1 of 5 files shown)


andrej (approved):

andrej:
This work will all need to be migrated to the new upstream MLIR-AIE kernels @efurst

PR #200, opened 2026-09-11
flm.GEMM: take M, K, N and the activation as runtime parameters

hunhoffe:
Moves flm.GEMM's shape out of the device configuration and into the instruction stream, so one xclbin serves every projection of a model instead of one per shape. All 30 E2B + E4B projection shapes now resolve to a single configuration, asserted by `test_one_xclbin_serves_every_shape`. M, K, N, the activation and the clamp bounds are runtime parameters.

The parameter buffer is 4 words rather than 10. A word costs ~2.06 µs of dispatch latency across the 32 cores, which is most of a short-prefill dispatch against a ~107 µs floor, so the clamp trio is sent only by a clamp-capable build and n_work/n_drain are derived on-core from a raw N word plus the tile's own column. Separately, split legs (K or N = 10240, where the row-block stride overflows the shim BD's 20-bit iteration step) now retire the oldest transfer as the next is issued instead of draining a whole window and then a whole column-block — worth up to -12.4% on those shapes.

The cost is memtile B residency, which sized the memtile buffer from k_iters and replayed it m_row_blocks times, putting both K and M into the configuration. Against the pre-RTP build, short prefill gets faster (M=256 median -5.5%) and long prefill pays for B's re-reads (M=1024 +13.1%, M=2048 +22.0%). This is a deliberate trade rather than a regression to fix before merge. Recovering it needs a replay count driven from the instruction stream; that is now reachable via `aiex.dma_channel_reset_for` in the pinned v1.4.3, but the buffer would still be sized from k_iters. Follow-up, not a blocker.

Accuracy is identical to pre-RTP.

## PR Merge Checklist

### DIFF
--- a/aie_kernels/generic/mm_fused.cc
+++ b/aie_kernels/generic/mm_fused.cc
@@ -35,8 +35,8 @@
 
 // Epilogue selection. 0 = none, 1 = gelu, 2 = silu, 3 = sigmoid, matching
 // Epilogue.mode in design.py.
-#ifndef MM_FUSED_EPILOGUE_MODE
-#define MM_FUSED_EPILOGUE_MODE 0
+#ifndef MM_FUSED_EPILOGUE_MODE_MASK
+#define MM_FUSED_EPILOGUE_MODE_MASK 0xF
 #endif
 #ifndef MM_FUSED_CLAMP
 #define MM_FUSED_CLAMP 0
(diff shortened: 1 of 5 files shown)


aie_kernels/generic/mm_fused.cc
```diff
@@ -35,18 +35,17 @@
 
 // Epilogue selection. 0 = none, 1 = gelu, 2 = silu, 3 = sigmoid, matching
 // Epilogue.mode in design.py.
-#ifndef MM_FUSED_EPILOGUE_MODE
-#define MM_FUSED_EPILOGUE_MODE 0
+#ifndef MM_FUSED_EPILOGUE_MODE_MASK
+#define MM_FUSED_EPILOGUE_MODE_MASK 0xF
 #endif
+// Whether a clamp is compiled in at all. Like the mode mask above this is a
```
hunhoffe: This seems to imply there are multiple builds per clamp. But doesn't this build include all the epilogue options? Isn't the logic to clamp or not static (and could be included in program memory?)

aie_kernels/generic/mm_fused.cc
```diff
@@ -144,47 +197,64 @@ void mm_fused_k_step(bfloat16 *a_buf, mm_fused_b_elem_t *b_buf, float *y_acc, in
 // Fusing the activation here is the point: the values are already in registers
```
hunhoffe: These few comment paragraphs seem excessive. Can you be more concise?

iron/operators/flm/gemm/design.py
```diff
@@ -97,6 +98,23 @@
 # (M_TILE * 256 * 4 = 65536 bytes) already fills the whole of L1, before A, B
 # or C are even counted, so no ct_max_k could ever make it fit.
 CT_MAX_K_FOR_N = {16: 16, 32: 32, 64: 128, 128: 32}
+# (tile_n, ct_max_k) pairs KNOWN TO COMPUTE CORRECTLY on hardware. The table
```
hunhoffe: TODO: can we dive deep? I don't like leaving artifacts from unsolved bugs in designs.

iron/operators/flm/gemm/op.py
```diff
@@ -143,16 +176,67 @@ def __post_init__(self):
...
+        dev = aie_utils.get_current_device().resolve().name
+        return (
+            f"tn{self.tile_n}_ck{CT_MAX_K_FOR_N[self.tile_n]}"
+            f"_ma{self.tile_ma}_mc{self.m_chunk}"
+            f"_em{self._epilogue_mask:x}_{self.rounding}"
+            f"_cl{self._clamp_capable}_{dev}"
+        )
+
+    @property
+    def config_name(self) -> str:
```
hunhoffe, line 191: is the onfig_tag actually reused? if not, can we combine functions instead of having mulitple helpers?

iron/operators/flm/gemm/op.py
```diff
@@ -201,31 +289,95 @@ def _link_file(self) -> str:
...
+        ):
+            # config_name, not name: this string reaches the design's
+            # link_with, so a shape in it would put the shape in the device
+            # configuration.
+            return f"{self.config_name}_kernels.a"
         return self._kernel_object
 
-    def get_mlir_artifact(self):
+    @property
+    def _reference_shape(self) -> tuple[int, int, int]:
```
hunhoffe, line 258: This... doesn't seem great. Can we use the runtime sequence generation via https://github.com/Xilinx/mlir-aie/blob/main/test/npu-xrt/matmul_whole_array_dynamic/whole_array_dynamic.py

requirements.txt
```diff
@@ -9,11 +9,16 @@
 # CUDA build served from PyPI. We therefore also pin torch to the "+cpu" local
 # version below, which is only available from the PyTorch CPU index.
 --index-url https://download.pytorch.org/whl/cpu
+--find-links https://github.com/Xilinx/mlir-aie/releases/expanded_assets/v1.4.3
 --find-links https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels-4
 --find-links https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
 --extra-index-url https://pypi.org/simple
 
-mlir_aie==1.4.3.dev85+gdf48abc
+# Tagged release wheels live under their own tag's asset page, not under
```
hunhoffe, line 17: We don't need a release comment here.

hunhoffe (commented):
More fine-tuning needed.

andrej (approved):

PR #204, opened 2026-09-16
Add optimized q4nx dequant operator

andrej:
This will need to push the MLIR-AIE pin once we have another nightly build with the kernel from https://github.com/Xilinx/mlir-aie/pull/3759


iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,266 @@
...
+        return ((cb // run_blocks) * period_blocks + cb % run_blocks) * cb_bytes
+
+    qw_bytes = qw_bytes_for(K, N, run_out_features, run_period_out_features)
+    out_blocks = K * N // BFP16_GROUP
+
+    # A shim tile's BD ids are shared across its channels, and finish() returns
+    # them to a compile-time allocator that does not check the transfer
+    # finished. A window must therefore fit the tile and be awaited before the
+    # next reprograms those ids. Both budgets count LIVE_WINDOWS windows; the
+    # fill is one descriptor per column block and outlives them.
```
andrej: Is this accurate? I think finish() does await, but only if the task it is waiting on has been configured with wait=True. This means it'll either lower to dma_await_task or dma_free_task (that's the non-waiting variant).

I think as it stands this comment is misleading.

iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,266 @@
...
+    k_window = min(
+        k_tiles,
+        SHIM_TASK_QUEUE // LIVE_WINDOWS,
+        (shim_bds - 1) // (LIVE_WINDOWS * HALVES),
+    )
+    if k_window < 1:
+        raise ValueError(
+            f"a shim tile with {shim_bds} descriptors and a queue of "
+            f"{SHIM_TASK_QUEUE} cannot hold one drain window"
+        )
```
andrej: I don't fully understand this math. Is k_window the size of the transfer? Do we ever want to do anything other than just overlap one transfer with another? Do we need this complex logic or could this just be expressed as splitting transfers in two so that one can run while we await the other?

iron/operators/flm/dequant/design.py
andrej: (Claude) You are right, and the comment was wrong. I checked `runtime.py`:

```python
if task.will_wait():
    group._actions.append((dma_await_task, [task.task]))
group._actions.append((dma_free_task, [task.task]))
```

So `wait=True` appends both an await and a free; `wait=False` appends only the free. `finish_task_group` then emits every await before every free.

The old comment described only the free half and implied `finish()` never awaits. In this design the drains all carry `wait=True`, so `finish()` genuinely awaits them.

Rewritten, and moved to the call site where it applies:

```python
# finish() awaits the transfers marked wait=True, then frees
# every descriptor in the group. Closing the previous window
# here overlaps its await with this one, which is already
# running.
```

and at the fill, which is the one transfer that does not wait:

```python
# The fill carries no wait. Freeing its descriptor is safe because
# a drain cannot complete before the cores consumed that fill.
```

Fixed in 2ae43a1.

andrej: (Claude) Correcting the replacement I quoted above — it was wrong twice over. "The fill carries no wait" states a property of fills in general, when it is an argument at one call site, and "carries no" is vague. Now:

```python
# This fill is not awaited; the drains are. A core reads the fill
# before it writes the output a drain takes, so a completed drain
# implies a completed fill.
```

Also swept the same pattern out of the rest of the change: `carries` became `has`, `uses` or `passes` in six places, and `The device configuration holds no K or N` became `No K or N reaches the device configuration`. 106044a.

iron/operators/flm/dequant/design.py
andrej: (Claude) You are right that it was more general than it needs to be, and the third term was dead.

To answer the questions:

**Is `k_window` the size of the transfer?** No. It is how many k-tiles' drains share one `TaskGroup`. Each k-tile contributes `HALVES` drains, one per k-half of the output slab, and each half has its own shim S2MM channel. Poor name; the comment now says what it counts.

**Do we ever want more than one transfer overlapping another?** No, and `LIVE_WINDOWS = 2` is exactly that: close the previous window only after the next is issued, so the await overlaps a running transfer.

**Was the logic needed?** Not as written. It was

```python
k_window = min(k_tiles, SHIM_TASK_QUEUE // LIVE_WINDOWS, (shim_bds - 1) // (LIVE_WINDOWS * HALVES))
```

which is `min(k_tiles, 2, 3)`. The descriptor term never binds — 15 // 4 = 3 against the queue's 4 // 2 = 2 — so it was dead generality dressed as a constraint. Now:

```python
k_window = min(k_tiles, SHIM_TASK_QUEUE // LIVE_WINDOWS)
live_bds = 1 + LIVE_WINDOWS * HALVES * k_window
if live_bds > shim_bds:
    raise ValueError(...)
```

One binding constraint in the formula, and the descriptor budget as a guard that states the real limit rather than silently shrinking the window.

**Could it just be two transfers, one running while the other is awaited?** That is the shape, and for every E2B projection it is literally what runs: `k_window` is 2 for any `k_tiles >= 2`, so a window is two k-tiles and two windows are live. What the expression still buys is the bound for large K — at K = 12288 there are 24 k-tiles, and they have to stream through that rotating pair rather than all be queued. If you would rather see `k_window = 2` with the queue arithmetic as a comment, say so; I kept it derived because `SHIM_TASK_QUEUE` and `LIVE_WINDOWS` are what actually set it.

Fixed in 2ae43a1.

andrej: (Claude) You were right, and further than I credited. I had conflated two quantities that both happened to be 2:

- `LIVE_WINDOWS = 2` — windows in flight. **This** is what overlaps the await with a running transfer.
- `k_window` — k-tiles *inside* one window. This does not affect overlap at all.

So `SHIM_TASK_QUEUE // LIVE_WINDOWS` was never answering "how many windows do I need". It was answering "how deep may I fill the channel queue", which is a throughput knob, not a requirement. With `k_window = 1` the overlap is identical: issue window i, await window i-1, 2 transfers outstanding per channel against a queue of 4.

So I measured it, interleaved, 15 samples per arm:

| K | N | k_window=1 | k_window=2 | delta |
|---|---|---|---|---|
| 12288 | 1536 | 1309.8 us | 1364.7 | **-4.0%** |
| 6144 | 1536 | 680.1 | 727.3 | **-6.5%** |
| 1536 | 12288 | 1233.3 | 1287.3 | **-4.2%** |

Same direction on all three. Deeper queueing costs.

So the answer is not `min(k_tiles, 2)` — it is 1, which deletes the concept. Gone: `k_window`, the `kb0` window loop, `SHIM_TASK_QUEUE`, the descriptor guard and the `get_target_model` import. The drain loop is now just `for kb in range(k_tiles)`, and what remains is one comment stating the resource use:

```python
# Drain windows in flight. A window retires once the next is issued, so its
# await overlaps the transfers of the next. One k-tile per window leaves two
# transfers outstanding per shim channel, and 5 of a shim tile's 16 buffer
# descriptors in use. Deeper queueing measured 4-6% slower.
LIVE_WINDOWS = 2
```

310 tests pass. 25e2467.

iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,241 @@
...
+    N,
+    tile_n=N_TILE,
+    qw_layout=QwLayout.FILE,
+    run_out_features=None,
+    run_period_out_features=None,
+    trace_size=0,
+):
+    """K in-features, N out-features. B reaches the GEMM as (K, N)."""
+    # The design generator round-trips its arguments, so a StrEnum arrives as a
+    # plain str and an identity test against it is silently false.
```
andrej:
```suggestion
```

iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,241 @@
...
+        f"q4nx_dequant_{get_kernel_dir(dev)}.o",
+        [qw_blk_ty, out_blk_ty],
+    )
+
+    def core_body(qw_in, out_of, k):
+        """One iteration is one q4nx block, and every block is identical work.
+
+        The loop takes no trip count and reads no runtime parameter, so no K
+        or N reaches the device configuration. The shim sequence bounds the
+        real work, and the loop then blocks on an empty input fifo.
```
andrej:
```suggestion
```

iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        for _ in range_(sys.maxsize):
```
andrej: Isn't there some sort of while_true=... flag on the Worker object or something in IRON so we could even skip this maxsize line?

andrej: (Claude) There is: `Worker(..., while_true: bool = True)`, and it defaults to True. `worker.py:294` wraps `core_fn` in `range_(sys.maxsize if self._while_true else 1)` itself, so the explicit loop nested a second one inside the first.

Removed, along with the `sys` and `range_` imports. The body is now:

```python
def core_body(qw_in, out_of, k):
    qw = qw_in.acquire(1)
    out = out_of.acquire(1)
    k(qw, out)
    qw_in.release(1)
    out_of.release(1)
```

Checked the emitted device body still carries `scf.for %arg0 = %c0 to %c9223372036854775807 step %c1` on all 32 cores, and 255 tests pass including the extensive K=12288 case. 8e2109c.

iron/operators/flm/dequant/design.py
andrej: (Claude) Taken, and the rest of that docstring went with the loop it described. The whole thing is now one sentence stating the reason rather than the mechanism:

```python
def core_body(qw_in, out_of, k):
    """One q4nx block, which is every block's work.

    Worker repeats this until reconfiguration, so it takes no trip count
    and reads no runtime parameter and no K or N reaches the device
    configuration.
    """
```

8e2109c.

iron/operators/flm/dequant/design.py
andrej: (Claude) Already gone. `QwLayout` was dropped entirely in 64cf35f2, along with that comment, the coercion it explained, `to_engine_order`, and the two tests that exercised the second order.

The reasoning: which block order the buffer holds is not testable here. The operator and the reference have to agree on which block covers which matrix region, and a test passes exactly when they do -- including when both are wrong. So it is a contract, not a result, and carrying two spellings of it plus a converter bought nothing. The order the fill reads linearly is now the contract, written down as layers 6 to 9 of the input table in README.md, and `dequantize()` indexes blocks that way.

That removed ~77 lines and every mention of "engine order" from the tree.

jgmelber:
Should this be included in the top level readme? Is that being kept up to date?

andrej:
@jgmelber Ah, good question. I can add it. I'll add the other new GEMM implementation to the top-level readme too as part of this PR since we haven't added that.

iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,214 @@
...
+from iron.common.device_utils import get_kernel_dir
+
+# q4nx block: 32 out-features x 256 in-features, 32 weights per scale and min.
+M_TILE, K_TILE, GROUP = 32, 256, 32
+BLOCK_BYTES = M_TILE * K_TILE * 5 // 8
+
+# flm.GEMM's B tiling. These must equal gemm/design.py's or the GEMM reads the
+# buffer in a different order than this design writes it.
+N_TILE, K_TILE_B, CT_K = 64, 512, 128
+S = T = 8
```
hunhoffe: Are S and T the same hardware derived things that can be derived by device type instead?

iron/operators/flm/dequant/design.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+            for r in range(ROWS)
+        ]
+
+    # A column block is read straight through. The block is split 10 x 512 so
+    # the innermost size stays inside the BD's field.
+    qw_sizes = [blocks_per_row, 2, BLOCK_BYTES // 512, 512]
+    qw_strides = [2 * BLOCK_BYTES, BLOCK_BYTES, 512, 1]
+
+    def qw_tap(cb):
```
hunhoffe: Lately, agents seem extremely happy to define a slightly absurd amount of helper functions. It might be worth double checking if these are really needed?

iron/tests/operators/flm_dequant_layout.py
```diff
@@ -0,0 +1,145 @@
...
+    S,
+    SLAB_BLOCKS,
+    T,
+)
+from iron.operators.flm.packing import f32_to_bfp16ebs8, pack_b
+
+# Every distinct (K in-features, N out-features) Gemma4 E2B needs, from
+# hidden_size 1536, intermediate_size 6144, DQ/DK/DV 4096/512/512 and the SWA
+# and skip variants.
+E2B_SHAPES = [
```
hunhoffe, line 35: Any chance we could add E4B too, to match testing for GEMM?

hunhoffe (approved):
Nice work!

iron/operators/flm/dequant/design.py
andrej: Refactored to get these out of the [GEMM instead](https://github.com/amd/IRON/blob/3873943a88057a526f40bcad7dc9371b37696a12/iron/operators/flm/gemm/design.py#L65), which is what this feeds, so the tile sizing must match

iron/operators/flm/dequant/design.py
andrej: good catch, inlined three functions with no clear standalone meaning and only one caller

iron/tests/operators/flm_dequant_layout.py
andrej: done, tested, works :D

PR #207, opened 2026-09-18
ci: bump llvm-aie alongside mlir_aie in the daily update workflow

atassis:
`requirements.txt` pins two halves of one toolchain -- `mlir_aie` and `llvm-aie` (Peano) -- but the
daily `update-mlir-aie` workflow only bumps the first, so the Peano pin moves only when edited by hand.

Checked against the live repo today: `requirements.txt` pins `llvm-aie==22.0.0.2026091701+773413fb`,
while the newest linux nightly wheel is `22.0.0.2026091801+7753da6c`. One day of drift, and nothing in
CI closes it.

## Added
- Two steps reading the current and latest `llvm-aie` versions.

## Changed
- The "create PR" condition now fires when either pin moves.
- The PR branch name folds in both versions. Keying it on `mlir_aie`'s tag alone collides across consecutive daily runs whenever only `llvm-aie` moved, and the existing branch-exists check turns that collision into a silently skipped Peano bump -- the common case, since Peano nightlies land far more often than `mlir_aie` releases.

## Removed
-

The latest nightly is read from the `nightly` **tag**, not `releases/latest`: the latter ranks by a
release's own publish date, which does not advance when new assets are attached to an existing
release, so it can report a wheel older than the one available.

Verified: the workflow YAML parses; `grep 'llvm-aie=='` extracts `22.0.0.2026091701+773413fb` from the
current `requirements.txt`; and running the new step's `gh api` query and `sed` against
`Xilinx/llvm-aie` returns `22.0.0.2026091801+7753da6c` -- it would open a bump PR today.

## PR Merge Checklist


.github/workflows/update-mlir-aie.yml
```diff
@@ -52,52 +80,71 @@ jobs:
           echo "tag=$latest_tag" >> $GITHUB_OUTPUT
 
       - name: Compare versions and create PR
-        if: steps.current_version.outputs.version != steps.latest_release.outputs.tag
+        if: steps.current_version.outputs.version != steps.latest_release.outputs.tag || steps.current_peano.outputs.raw_version != steps.latest_peano.outputs.raw_version
```
andrej: I think if we trigger this on every nightly peano release we'll basically have a PR open for a peano bump every night. Can we leave this condition unchanged, so it only triggers alongside MLIR-AIE release?

Or even better, just pick the peano release pinned inside MLIR-AIE, because we know those are matched?

andrej (changes requested):

andrej (approved):

PR #208, opened 2026-09-18
swiglu_prefill: add b_col_maj, the weight layout the GEMMs already take

atassis:
`GEMM` already takes `b_col_maj` to read its `B` operand as `(N, K)` instead of `(K, N)`.
`SwiGLUPrefill` builds three GEMMs and does not expose it, so its weights can only be stored one way.

That matters for a rail sharing one weight arena between prefill and decode: the decode-side GEMV
reads `(N, K)`, so today prefill needs a transposed second copy of every projection weight.

## Added
- `b_col_maj` on `SwiGLUPrefill`, threaded to all three GEMMs.
- A `b_col_maj=True` case on the existing device test, and a construction-only test asserting the flag reaches every GEMM in the runlist.

## Changed
- The design name gains a `_bc` suffix under `b_col_maj`, so the two layouts cache as separate artifacts rather than colliding.

## Removed
-

Verified: `pytest -k weight_layout_reaches_both_gemms` passes (10 cases, no device); `black --check`
clean. The parametrized **device** case is not run here -- no NPU on this checkout -- so the
`b_col_maj=True` numerical path needs a CI device leg or a maintainer run.

## PR Merge Checklist


iron/operators/swiglu_prefill/op.py
```diff
@@ -17,15 +17,27 @@ class SwiGLUPrefill(OperatorSequence):
     tokens. Runtime buffers (via ``get_callable().get_buffer(name)``): input
     ``in``; persistent weight scratch ``w_gate`` / ``w_up`` / ``w_down``;
     output ``out``.
+
+    ``b_col_maj`` selects the layout all three weights are stored in, ``(K, N)``
+    when False and ``(N, K)`` when True. It is the layout the decode-side GEMV
+    reads, so a rail sharing one weight arena between prefill and decode sets it
```
andrej: What's a "rail"?

andrej (approved):

PR #203, opened 2026-09-15
aie_kernels: source all C++ kernels from mlir-aie, remove in-tree copies

efurst:
IRON no longer hosts C++ compute kernels. Every operator now sources its kernel
  `.cc` from the installed `mlir_aie` package via `AIEContext.kernels_dir`
  (`root_path()/include/aie_kernels`), and the in-tree `aie_kernels/` directory is
  removed. This is the IRON half of the kernel consolidation started by
  [mlir-aie#3599](https://github.com/Xilinx/mlir-aie/pull/3599).

  > **Depends on [Xilinx/mlir-aie#3740](https://github.com/Xilinx/mlir-aie/pull/3740).**
  > Several kernels IRON sourced in-tree needed additions upstream before IRON
  > could consume them from the package (runtime-size eltwise/activation entry
  > points, runtime-epsilon `rms_norm`, device-generic `rope`, the fused-GEMM
  > `mm_fused` cluster, and `aie2` `mask_bf16`/`layer_norm`). This PR builds
  > against a wheel that contains those additions.
  > 
  > **Merge blockers, in order:** (1) #3740 merges → (2) a new `mlir_aie` wheel
  > publishes → (3) bump the `requirements.txt` pin to it. Until then the default
  > build resolves to a wheel without the new symbols; set
  > `IRON_AIE_KERNELS_DIR=<local mlir-aie>/aie_kernels` to build against a local
  > checkout. 

  ## Added
  - `AIEContext.kernels_dir` — resolves kernel sources to the installed
    `mlir_aie` package (`aie.utils.config.root_path()/include/aie_kernels`).
  - `IRON_AIE_KERNELS_DIR` env override on `kernels_dir`, to point IRON at a local
    mlir-aie checkout for kernel development. _Happy to remove this if it isn't wanted._
    
  ## Changed
  - Every operator's `get_kernel_artifacts()` (and the stream-op / flm-gemm
    kernel sourcing) now reads from `context.kernels_dir` instead of the in-tree
    `aie_kernels/`. Where mlir-aie renamed or re-homed a symbol, the operator's
    `Kernel(...)` binding was updated to match (e.g. `convert_copy_f32_to_bf16` →
    `cast_f32_bf16_row`, `expand_int4_to_bfloat16` → `expand_uint4_to_bfloat16`,
    the `_size`/`_eps` runtime variants, and `rope`'s two-symbol scheme).
  - README / AGENTS.md kernel references point at the mlir-aie kernel library.

  ## Removed
[...]

PR #210, opened 2026-09-18
stream: keep the ONNX registry off the operator import path

atassis:
`iron/common/stream/__init__.py` states the rule -- its submodules need `onnx`/`pyyaml`, "so importing
an operator must not pull them in" -- and `requirements_stream.txt` promises the stream operator's
test skips itself when stream-dse is absent. Neither holds.

`iron/operators/__init__.py` re-exports `SwiGLUPrefillStream`; `swiglu_prefill_stream/op.py` imports
`ELTWISE_MUL`, `GEMM` and `SILU` from `iron.common.stream.ops`; and that module imports `onnx` and
`onnxscript` at module scope. So **a core install cannot import `iron.operators` at all** -- a
`ModuleNotFoundError` at collection, which takes out every operator's tests, not just the one operator
the optional dependency belongs to.

## Added
- `iron/common/stream/kernels.py`, holding the `StreamKernel` definitions and operand layouts.
- `iron/tests/core_install.py`, pinning the contract: a child interpreter with `onnx`/`onnxscript` hidden from `sys.meta_path` must still import `iron.operators` and `iron.common.stream.kernels`.

## Changed
- `ops.py` keeps only the registry it is named for -- the private-domain schemas, the translation table, and `TORCH_OPS` -- and imports the kernels it references.
- `swiglu_prefill_stream/op.py` and `iron/tests/stream/kernel_layouts.py` take their names from the new module.

## Removed
-

Splitting rather than deferring the import into the functions that use it is deliberate: a module that
does not import `onnx` cannot regain the dependency by accident.

Verified on `27cf75d`: `pytest iron/tests/core_install.py` -> 20 passed; `kernel_layouts.py` skips as
designed without stream-dse; `TORCH_OPS` still resolves 3 ops and `translation_table()` 3 entries;
`black --check` clean.

Note on rebasing: #203 changed the moved block (the `bfp16_mmul` MAC-row split and `_gemm_artifacts`'
`base_dir` -> `kernels_dir` rename). `kernels.py` carries `devel`'s current version of that block moved
verbatim, not the older copy this branch was first written against.

## PR Merge Checklist


andrej:
@asyms As a sanity check, does this make sense?

atassis:
This work was done before #150 landed, which removed re-export. Some redesign also happened at that time, so right now this is mostly no-op.
The only thing left is a preference rather than a defect: `ops.py` still requires `onnx` from anyone who wants a `StreamKernel`, which is not an ONNX concept.

PR #211, opened 2026-09-18
shim BD: derive the granule from dtype, and share one run-split helper

atassis:
`gemv` and `repeat` both split a contiguous run into shim BD wrap dimensions, each hand-rolling the
same invariant. `gemv` also carried a `FIXME` asking for the bounds to come from the target model, and
a granule constant that was bf16-only:

```python
GRAN_ELEMS = 2  # 4-byte shim granularity / 2-byte bf16 element
```

That is wrong for every other dtype it accepts. The shim's address-generation granularity is 32 bits,
so the granule is 4 elements for int8 and 1 for f32 -- an element-unit bound derived from 2 is 2x too
strict for int8 and 2x too loose for f32.

## Added
- `iron/common/shim_bd.py` with one `split_run` helper, and its tests.
- `_shim_gran_elems(dtype)`, deriving the granule from the memref's element type, mirroring what `verifyStridesWraps` computes from `getAddressGenGranularity()`. It asserts the element width divides the granularity rather than silently truncating.

## Changed
- `gemv` and `repeat` both use the shared helper.
- The bound comment now records that these are the **ShimNOC** numbers specifically; MemTile and CoreTile have different wrap/step widths, so reusing them elsewhere would silently under- or over-shoot.

## Removed
- `gemv`'s local `split_run`, `MAX_WRAP`/`GRAN_ELEMS`/`MAX_STRIDE`, and the `FIXME`.

The bounds stay hard-coded, with a comment saying why: `getDmaBdWrapBits`/`getDmaBdStepBits` have no
Python binding, and `getAddressGenGranularity` has a CAPI entry but is not bound into
`aie.dialects.aie.AIETargetModel`.

Verified on `27cf75d`: `pytest iron/common/test_shim_bd.py iron/operators/gemv/test.py` -> 60 passed
for the shim-BD and granule selections, covering bf16/int8/f32 granules and the split invariant.
`black --check` clean. Pure arithmetic; no device. The gemv/repeat device paths are unchanged in shape
but not re-run here.

## PR Merge Checklist


andrej (commented):
Could this be an upstream change to MLIR-AIE instead, where we can access/expose the required fields from the TargetModel? And make IRON a user of it only.

atassis:
@andrej really the job was done in xilinx/mlir-aie#3728. So after that lands- I can wire it into the IRON later

andrej:
That's even better. Please ping here when upstream lands, and we can update this PR to make use of those changes. I'll hold off until merging until then and mark it draft until then.

PR #212, opened 2026-09-18
compilation: byte-address the fused arena, type buffers by operator

atassis:
`fuse_mlir` types the consolidated arena bf16 and divides every byte offset and length by
`np.dtype(ml_dtypes.bfloat16).itemsize`:

```python
buf_dtype = np.dtype[ml_dtypes.bfloat16]   # TODO: support for other data types
itemsize = np.dtype(ml_dtypes.bfloat16).itemsize
```

`subbuffer_layout` already stores **byte** offsets and lengths, so that divisor is correct only while
every buffer in the sequence is bf16. Any other dtype has its element count silently halved -- a
576-byte uint8 buffer (Dequant's packed input) is declared as 288 elements.

## Added
- `element_size_bytes(ty)`, the single owner of this seam's byte<->element rate. Sub-byte types are refused rather than rounded: packed nibbles plus interleaved scales are not a whole number of elements of any single type, so such a buffer must declare the byte type it occupies.
- `iron/tests/infrastructure/element_size.py`, covering integer widths, the float types, i4/i1 refusal, and that the error names the unsupported type.

## Changed
- The arena is typed `int8`, so byte offsets and lengths pass through unscaled; each operator's view is typed from its own element type.

## Removed
- The `ml_dtypes` import, whose last use byte-addressing removes.

Verified on `27cf75d`: `pytest iron/tests/infrastructure/element_size.py` -> 70 passed, device-free;
`black --check` clean; `ruff check` reports no diagnostic that `devel` does not already report on this
file.

## PR Merge Checklist


iron/common/compilation/sequence.py
```diff
@@ -31,6 +30,33 @@
...
+    if isinstance(ty, ir.IntegerType):
+        bits = ty.width
+    elif isinstance(ty, (ir.BF16Type, ir.F16Type)):
+        bits = 16
+    elif isinstance(ty, ir.F32Type):
+        bits = 32
+    elif isinstance(ty, ir.F64Type):
+        bits = 64
+    else:
+        raise TypeError(f"fused arena: unsupported element type {ty}")
```
andrej: Just to double check -- is there truly nothing on ir.F...Type that returns the bit-width directly? If we have to do it this way, by enumerating all types, so be it, but I just wanted to ask to be sure.

PR #197, opened 2026-09-11
Add artifact fingerprinting and improve validation

ak495867:
Introduce deterministic build-product fingerprinting and tooling-aware cache invalidation in iron/common/compilation/base.py: DataFlowTier enum, toolchain fingerprint, .meta sidecars, fingerprint() logic, and record_fingerprint() calls to record/revalidate artifacts. Switch availability checks to compare fingerprints instead of mtimes. Record fingerprints after kernel/xclbin/ELF production. Replace many in-code assert checks across operators with user-friendly ValueError messages and clearer diagnostics (gemm, gemv, gemv/ref, gemv/design, mha, rope, strided_copy, strided_copy/design, mem_copy, softmax, rms_norm, etc.). Misc cleanup: remove unused imports, tidy prints, adjust kernel name usage, and small import reorders (pretty_aggregate, conftest, llama harness). These changes improve build caching correctness and make runtime parameter validation more explicit and debuggable.

Describe the intent of your PR here.

## Added
-

## Changed
-

## Removed
-

## PR Merge Checklist

1. [ ] The PR is rebased on the latest `devel` commit and pointing to `devel`.
2. [ ] Your PR has been reviewed and approved.
3. [ ] All checks are passing.

PR #213, opened 2026-09-18
compilation: content-address the build cache instead of comparing mtimes

atassis:
`CompilationArtifact.is_available_in_filesystem()` decides cache hits by comparing mtimes. There is no
content hash anywhere in that check, so any input that changes an artifact without touching a source
file's timestamp is invisible to it.

I hit this as a live bug on 2026-09-12: rebuilding row-batched vs runtime-K scores-GEMV arms into a
shared build cache silently linked the **wrong compiled object**, because the flag that changes the
object never enters the comparison. It is the same class as the arch collision #166 patched around --
that fix namespaced kernel objects per-arch rather than fixing the comparison, so the next axis nobody
thought to namespace reopens it.

The key is computed from the compiler's **own `-MD`/`-MF` depfile** rather than a hand-enumerated
input list. Every prior instance of this defect class -- this one, `ExternalFunction._content_digest`,
`@iron.jit`'s original `_hash.py` -- was a caller that forgot an include, so enumerating inputs by hand
is the thing to avoid, not to redo more carefully.

## Added
- `iron/common/cache.py`: `content_key()` / `ContentStore`.
- A sidecar manifest per kernel object recording the key that produced what is on disk, checked against a fresh recompute of (depfile inputs, `extra_flags`/`rename_symbols`/`prefix_symbols`/`use_chess`, Peano's own `--version` identity). `KernelCompilationRule` writes it after every successful compile.
- Three test files, each with hand-built fixtures, a real-recording-path exercise, and a `content_key`-sabotage proof that the test is not vacuous; plus a Chess-is-never-recorded test.

## Changed
- `KernelObjectArtifact.is_available_in_filesystem()` is content-addressed.
- `PythonGeneratedMLIRArtifact` keys on `(source_path, fn_name, args, kwargs)`. Two `DesignGenerator` calls differing only in a kwarg previously shared one cached "available" MLIR file. `args`/`kwargs` enter the key only when every value is a JSON scalar (or a list/dict of those); anything else **fails closed**, since a plain object's default `repr()` embeds its memory address and would make the key silently non-deterministic.
- `FullElfArtifact` / `XclbinArtifact` / `InstsBinArtifact` key on `self.dependencies` -- every linked kernel object and archive, which they already list -- instead of re-deriving aiecc's `link_with` regex scan as a second, weaker copy.

## Removed
-

**Peano only, deliberately.** No construction site in this tree threads a `use_chess` choice through to
these artifact classes, so a Chess build has no toolchain fingerprint to check a manifest against.
Recording is skipped under `use_chess=True`, leaving such an artifact conservatively mtime-only rather
than checked against a Peano fingerprint it was never built with. `compile_mlir_module`'s own cache
still makes a repeat Chess compile fast, just not skipped at this outer layer.

Two known gaps, stated rather than papered over: this does not see transitive imports a generator's
module makes -- a Python callable has no depfile equivalent to enumerate them, unlike the C++ side --
and `ExternalFunction._content_digest` and `@iron.jit`'s `_hash.py` are untouched, since they live in
mlir-aie and whether that repo should depend on this package for the shared primitive is an open
design question.

Verified rebased on `27cf75d`: `pytest iron/tests/compilation/` -> **190 passed, 0 failed** (was 95 on
`devel`, so no regression). `black --check` clean.

## PR Merge Checklist


andrej:
Unfortunately we are not accepting modifications to the compile infrastructure here in IRON, because we consider it deprecated in favor of the @iron.jit decorators in MLIR-AIE. Our plan is to drop the custom compile infrastructure in this repository in favor of what we have in MLIR-AIE, though the timeline is currently unknown. There already has been significant effort put into cache invalidation for that implementation, and we fear that adding more functionality to IRON's compile infrastructure will make deduplication harder.

We are very interested in contributions to (a) the migration to @iron.jit in this repository, and (b) improvements, to the corresponding upstream mechanisms in MLIR-AIE.

See also #197.

